PR #276, opened 2026-05-24
Replace Cecil-based ReSharper annotation export with a Roslyn exporter

### LINKED ISSUE #275: Replace Cecil-based ReSharper annotation export with a Roslyn source-reading exporter
### Reference version

`2.0.134-preview`

### Background and motivation

The `JetBrainsAnnotations` SDK module lets consumer projects (`UseJetBrainsAnnotations=true`) ship ReSharper code annotations to _their_ downstream consumers without leaving JetBrains attributes in the compiled assembly. It currently does this via `ReSharper.ExportAnnotations.Task`, which depends on Mono.Cecil and on a second helper project, `Cecil.XmlDocNames`.

This pipeline has three structural problems:

1. **Cecil cannot reliably strip the attributes from the assembly.** The workaround is to build with `JETBRAINS_ANNOTATIONS` defined, export annotations, then rebuild _without_ the symbol — a build-twice dance that hooks the standard `Build` task with no future-proofing guarantee.
2. **Mono.Cecil is an unattractive long-term dependency** — sparse updates, a large backlog of open issues.
3. The pipeline is powered by two external repos we'd like to be free to retire.

Since this module was first written, `JetBrains.Annotations.Sources` has appeared: it ships the annotation attributes as `internal` types decorated with `[Conditional("JETBRAINS_ANNOTATIONS")]`. With the symbol undefined (the normal build), the compiler never emits the attribute _usages_ into metadata — so the "strip" step Cecil was failing at happens for free, at compile time. The leftover unused internal attribute types are dropped by the trimmer/ILC, leaving IL _and_ AOT binaries clean.

That removes the _reason_ for the build-twice hack and for reading/rewriting the assembly. The only remaining job — turning source annotations into ReSharper external-annotations XML — is something Roslyn does natively: `GetDocumentationCommentId()` produces the exact member IDs that `Cecil.XmlDocNames` was hand-rolling.

Affected project: **Buildvana SDK** (the `JetBrainsAnnotations` module). This issue is scoped to the Buildvana-side replacement only.

### Proposed enhancement

For **C# projects** with `UseJetBrainsAnnotations=true`, Buildvana SDK should:

- build normally, with `JETBRAINS_ANNOTATIONS` **undefined**, so no JetBrains attributes appear in the compiled assembly (clean IL and AOT output);
- emit a `{AssemblyName}.ExternalAnnotations.xml` and pack it **alongside `{AssemblyName}.dll`** (`lib/<tfm>/`), one file per target framework, which is all ReSharper's discovery requires;
- do all of the above **without Mono.Cecil, without a second build pass, and without hooking the `Build` task**.

VB (`.vbproj`) support is **dropped** as part of this work (see Risks/Acceptance criteria): the module will force `UseJetBrainsAnnotations=false` for non-C# project types.

**Acceptance criteria**

- [ ] C# projects with `UseJetBrainsAnnotations=true` produce `{AssemblyName}.ExternalAnnotations.xml` next to the assembly, per TFM, with member IDs from `GetDocumentationCommentId()`.
- [ ] Compiled assemblies contain no JetBrains attribute metadata (verified for IL; confirmed clean under AOT/trim).
- [ ] No dependency on Mono.Cecil, `ReSharper.ExportAnnotations.Task`, or `Cecil.XmlDocNames`; no second build pass; no `Build`-task hooking.
- [ ] Annotation **constructor arguments and named properties** round-trip into the XML, covered by tests (see Risks).
- [ ] `.vbproj` opt-in is disabled.
- [ ] Changelog updated under `## Unreleased changes` → `### Changes to existing features`, with `**BREAKING CHANGE**` for the VB drop.

### Implementation proposals

1. **Switch the injected reference.** In `JetBrainsAnnotations/Module.targets`, replace the compile-only compiled `JetBrains.Annotations` reference with `JetBrains.Annotations.Sources` (already `developmentDependency=true`, won't flow transitively). Remove the `ReSharper.ExportAnnotations.Task` reference and the `CheckForJetBrainsAnnotationsPackageReference` / `UpdateJetBrainsAnnotationsPackageReferencee` workarounds. Narrow the opt-in coalescing to `.csproj` only.

2. **Add a Roslyn-based exporter** that reads JetBrains annotation attributes from **source** (mandatory — they're absent from post-build metadata; see Additional information) and emits the XML, member IDs via `ISymbol.GetDocumentationCommentId()`.
   - **Primary: a standalone MSBuild task** in `Buildvana.Sdk.Tasks` (net10) referencing `Microsoft.CodeAnalysis.CSharp`, building a `Compilation` from `@(Compile)` + `@(ReferencePathWithRefAssemblies)`, mirroring the real compile's `LangVersion` and `DefineConstants` (minus `JETBRAINS_ANNOTATIONS`). Because MSBuild now runs tasks **out of process**, the task can load its own Roslyn without colliding with MSBuild's — the earlier assembly-load concern is moot, provided we don't conflict with what the task host itself needs.
   - **Rejected: incremental source generator.** Reads the live compilation with no re-parse, but generators are meant to emit source, not files; writing an artifact during design-time/IDE builds is wrong. Noted for completeness.
   - Gate the exporter to pack time (or an incremental target with proper inputs/outputs), since in-solution ReSharper reads annotations from source directly and only the packaged consumer scenario needs the XML.

3. **Update the changelog** per the acceptance criteria.

### Usage examples

Consumer project, unchanged opt-in:

```xml
<PropertyGroup>
  <UseJetBrainsAnnotations>true</UseJetBrainsAnnotations>
</PropertyGroup>
```

Annotated source:

```csharp
public string Bar([NotNull] string baz) => baz;
```

Packaged `DemoLib.ExternalAnnotations.xml` (shape validated by the spike):

```xml
<assembly name="DemoLib">
  <member name="M:Demo.Foo.Bar(System.String)">
    <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
    <parameter name="baz">
      <attribute ctor="M:JetBrains.Annotations.NotNullAttribute.#ctor" />
    </parameter>
  </member>
</assembly>
```

The consumer's compiled assembly contains **no** JetBrains attribute metadata; its AOT image carries none either.

### Risks

- **Attributes with constructor arguments / named properties** (`StringFormatMethod`, `ContractAnnotation`, `MeansImplicitUse`, `BaseTypeRequired`, …). The XML format encodes `<argument>` values, not just `#ctor`. The spike only exercised a parameterless attribute; faithfully serializing ctor args and named values is the main correctness surface — hence the dedicated tested acceptance criterion.
- **Parse-option drift.** The task's reconstructed compilation must track the real compile's define constants, lang version, and references, or symbol/attribute resolution will diverge.
- **VB drop is a breaking change** for any `.vbproj` consumer currently relying on the module. Acceptable per scope decision; called out in the changelog.

### Additional information

Validated with a Roslyn 5.3.0 spike (symbol **undefined**, mirroring `JetBrains.Annotations.Sources`' internal + `[Conditional]` attribute):

| Check | Result |
|---|---|
| `GetAttributes()` on **source** symbols | `[NotNullAttribute]` on method and parameter — visible to a generator/analyzer/task, so we read from the semantic model directly |
| `GetAttributes()` on **metadata** symbols (post-emit) | empty — usages stripped from the assembly → clean IL and AOT |
| Attribute **type** in metadata | present but unreferenced → trimmer/ILC drops it |
| `GetDocumentationCommentId()` | `M:Demo.Foo.Bar(System.String)`, `T:Demo.Foo` — replaces `Cecil.XmlDocNames` |

Key consequence: because the attributes are **gone from metadata after build**, the exporter _must_ read from source — which is precisely what lets us drop both the build-twice pass and Cecil's assembly reading/rewriting.

rdeago:
## Summary

Reworks the `JetBrainsAnnotations` SDK module to export ReSharper external annotations by reading `JetBrains.Annotations` attributes from **source** via Roslyn, instead of reading/stripping the compiled assembly with Mono.Cecil.

- New `ExportJetBrainsAnnotations` task (in `Buildvana.Sdk.Tasks`) reconstructs a `CSharpCompilation` from `@(Compile)` + `@(ReferencePathWithRefAssemblies)`, mirroring the project's `LangVersion`/`DefineConstants`, and emits `{AssemblyName}.ExternalAnnotations.xml` with member IDs from `GetDocumentationCommentId()`.
- The export runs after the (per-TFM) build and is incremental; the file is packed next to the assembly in `lib/<tfm>`, one per target framework. No second build pass and no `Build`-task hooking.
- The core exporter logic lives in an MSBuild-free internal class so it can be unit-tested in a later PR.
- Roslyn is referenced compile-only (`ExcludeAssets="Runtime"`): the MSBuild task host always provides it at run time, so nothing extra is bundled into the SDK package. Verified end-to-end against a sample consumer.

### Breaking changes

- **The `UseJetBrainsAnnotations` property is removed, and the SDK no longer injects any JetBrains annotations package.** Opt in with the new boolean `ExportJetBrainsAnnotations` (default `false`) and reference an annotations source yourself — the compiled `JetBrains.Annotations` package, `JetBrains.Annotations.Sources`, or your own attributes in the `JetBrains.Annotations` namespace.
- **Visual Basic (`.vbproj`) projects are no longer supported by the module.** `ExportJetBrainsAnnotations` is forced to `false` for any non-C# project, because the exporter reads C# source directly.

### Notes

- Annotation **constructor arguments** are serialized (positional `<argument>` elements); named properties are initialized from constructor arguments and need no separate representation. This matches ReSharper's external-annotations format and the prior exporter.
- `JETBRAINS_ANNOTATIONS` is left entirely to the user: the exporter mirrors the real compile's `DefineConstants` and reads annotations through the semantic model, so the XML is produced identically whether or not the symbol is defined.
- New diagnostic `BVSDK1300` (export failure) is documented in `docs/Diagnostics.md`.
- Tests are deferred to a later PR, as per internal discussions.

## Test plan

- [x] Full solution builds with zero warnings/errors (warnings-as-errors).
- [x] ReSharper `inspectcode --severity=WARNING`: 0 results.
- [x] End-to-end against a sample C# library consuming the locally-packed SDK with `ExportJetBrainsAnnotations=true`: build produces a correct `*.ExternalAnnotations.xml` (parameterless attributes, constructor arguments with XML escaping, parameter and method annotations), and `dotnet pack` places it at `lib/<tfm>/` next to the assembly.

Closes #275


rdeago:

rdeago:
Thanks for the thorough review. Addressed point by point:

**1. Member-level visibility** — Fixed in faa7119. `TypeToXml` now skips members that aren't externally visible, using the same `Public | Protected | ProtectedOrInternal` predicate that gates types (extracted into a shared overload so the two can't drift). Note this also excludes explicit interface implementations (they're `Private`) — that's intentional: a consumer uses the interface, so that's where ReSharper looks for annotations, not the implementation.

**2. `AllowUnsafe`** — Fixed in a5ba739. Pass `allowUnsafe: true` so the reconstructed compilation matches consumer projects with `AllowUnsafeBlocks`.

**3. Diagnostics.md alignment** — Fixed in 4dd50cc. Repadded the `Error` cell to match the other error rows.

**4. `FormatArgument` fallback** — Left as-is, agreeing it's fine as a defensive fallback.

@claude please review again — the latest changes are pushed.

PR #287, opened 2026-05-28
Replace bv's logging with Buildvana.Core.ConsoleOutput

### LINKED ISSUE #286: Replace bv's Microsoft.Extensions.Logging usage with a purpose-built console-output library
### Reference version

`2.0-preview` (current `main`).

### Background and motivation

`bv` uses Microsoft.Extensions.Logging purely to write to the console. That's a mismatch: a CLI tool's console output is **UI for a human watching now**, not **diagnostic telemetry for a sink**. Five concrete problems follow from it:

1. **`dotnet`/MSBuild output is swallowed** unless the child process fails; on failure only the first 20 + last 20 lines survive, inside an exception message. A build with 50 errors loses most of them.
2. **Logger categories are noise** — a user doesn't care that `Buildvana.Tool.Services.DotNetService` emitted a line.
3. **Class-name categories leak internals and create rigidity** — renaming a class (a non-user-facing refactor) silently changes user-facing output, and external tooling could come to depend on the category text.
4. **`bv` never logs to a file**, so the logging pipeline is conceptually dead weight.
5. **No formalized grouping.** A "begin → detail → done" activity pattern deserves to be a type. We already fake it: the `"Git: "` message prefixes and the `CreateLogger("Release")` / `CreateLogger("Clean")` string categories are hand-rolled activity labels.

This affects every `bv` user; it's not specific to any project type or toolchain.

### Proposed enhancement

`bv`'s output becomes purpose-built console UI:

- Messages render as clean, color-coded lines — **no class-name category prefix** (each level keeps a short, color-coded label).
- **`dotnet`/MSBuild output streams through live** instead of being hidden until failure; the head/tail buffer is retained only as the quiet-mode failure tail.
- **Verbosity is unchanged** (`--verbosity quiet|minimal|normal|detailed|diagnostic`).
- **Activity grouping** (begin → detail lines → outcome) becomes a first-class concept, replacing the ad-hoc prefixes/string categories.

### Implementation proposals

- New host-agnostic library **`Buildvana.Core.ConsoleOutput`**; contracts (`IReporter`, `IActivityScope`, `Verbosity`, `MessageLevel`, `NullReporter`) in the shared `Buildvana.Core.Abstractions`; concrete `ConsoleReporter` **writing via `System.Console` (BCL-only — no rendering dependency)**. The level→color intent in the existing `SpectreLogger` ports over (mapped to `ConsoleColor`), minus the category.
- Stream child output: extend `IProcessRunner` to fan out **stdout and stderr** live (today only stdout has an unused hook, and `DotNetService` never wires it). Keep `HeadTailPipeTarget` as the quiet-mode failure tail.
- Remove Microsoft.Extensions.Logging from `Buildvana.Tool`; delete the `SpectreLogger`/`SpectreLoggerProvider` bridge; migrate the ~17 `ILogger` call sites to `IReporter`.
- **Out of scope:** `Buildvana.Sdk.Tasks`' `ILogger`→`TaskLoggingHelper` bridge (different host: MSBuild), and `bv`'s help rendering (`BvHelpRenderer`), which keeps using Spectre.Console.

### Usage examples

Before:

```
info: Buildvana.Tool.Services.DotNetService[0]
      Building solution (restore = False)...
```

(…and on a failed build, `dotnet`'s 50 errors collapse to a 20+20-line excerpt inside an exception.)

After:

```
Building (Release)
  <dotnet/MSBuild output streams here, live and in full>
  done (3.2s)
```

### Risks

- **User-visible output format change.** `bv`'s output isn't a documented contract, but anything scraping it would see a difference. This is a **non-breaking** change to "Changes to existing features" in the changelog.
- **Dependency changes:** `Microsoft.Extensions.Logging`(`.Abstractions`) dropped from the Tool; `Spectre.Console` stays for help rendering only; the new `Buildvana.Core.ConsoleOutput` library is BCL-only (no third-party rendering dependency).
- Verbosity semantics are preserved, so that surface is low-risk.

### Additional information

- Acceptance criteria include the non-breaking changelog entry above and a clean `dotnet bv pack` + ReSharper `inspectcode` run.

rdeago:
## Summary
- Adds `Buildvana.Core.ConsoleOutput`: a `System.Console`-backed `IReporter` (contracts in `Buildvana.Core.Abstractions`, BCL-only) that replaces `bv`'s misuse of `Microsoft.Extensions.Logging` for console output.
- Messages now render as clean, color-coded lines with no class-name category; `dotnet`/MSBuild output streams through **live** (stdout→stdout, stderr→stderr) instead of being hidden unless the build fails. Work is grouped into depth-tagged **activities** (`[1] Build` … `[1] Build: done (3.2s)`).
- Removes the `SpectreLogger` bridge and `Microsoft.Extensions.Logging`(`.Abstractions`) from the tool; `Buildvana.Sdk.Tasks` keeps its MSBuild bridge with an explicit `M.E.Logging.Abstractions` reference.

## Notable design choices (reviewer context)
- **Palette:** foreground-only, theme-adaptive — `error:` red, `warning:` yellow, `info:`/`detail:`/`trace:` uncolored; labels ragged (single space). `NO_COLOR` honored; `--color` overrides it.
- **Activities:** `IActivityScope.Complete()` gates the outcome line (no "done" on exception); both the starting and outcome line render at Info; an instance stack supplies the `[depth]` tag.
- **`GitService` was not given an activity** (its logging is all-`Detail` discovery + silent ops); a "Git" activity likely belongs at the callers later.
- **Ctrl-C** now reports via `error:` (always visible) rather than the old yellow line.

## Additional changes (not in the issue)
- Improved overall handling of `dotnet` invocations in `DotNetService`.
- Removed `dotnet` parameters of non-general use (they will be added in this repo's `buildvana.jsonc` file when it is created).
- Added DI support for `Lazy<T>`: an injected `Lazy<T>` will resolve `T` from the service provider when evaluated. This lets us inject `Lazy<NuGetPushConfiguration>` into `DotNetService` instead of the service provider itself.

## Test plan
- [x] `dotnet bv pack` — 0 errors, 0 warnings, tests pass, artifacts produced
- [x] ReSharper `inspectcode --severity=WARNING` — 0 results
- [x] Ran the locally-built tool against a throwaway repo: verified activity headers + depth, live MSBuild streaming, and the `error:` path

Closes #286


rdeago:

rdeago:
@claude thanks for the review!

> **Activity "done" at Detail vs "begin" at Info**

They're both at Info now.

> **`DotNetService.cs:256` — declaration exceeds 120 chars**

Fixed.

> **`NO_COLOR` spec compliance** — The check `is { Length: > 0 }` silently ignores `NO_COLOR=` (set but empty). The [NO_COLOR spec](https://no-color.org/) says to disable color "regardless of value".

The spec says "when present _and not an empty string_ (regardless of its value)" (emphasis mine). Our check is fully compliant.

> **`NullReporter.Verbosity = Quiet`** — `IsEnabled(Error)` returns `true` for `Quiet` (0 ≤ 0), so error-level `CompositeFormat` messages on the null reporter still incur string formatting before being silently discarded.

There's no `None` verbosity by design choice; `Quiet` is the verbosity level that best approaches what `NullReporter` actually does. More than a handful error messages in the same run are an extreme improbability, so further optimization would be pointless.

I have also added other changes, not in the original scope of the PR. Please review again.

rdeago:
Thanks again @claude!

> **`ChildOutput` / `ChildError` verbosity filter is inverted** — `ConsoleReporter.cs` lines 74–83 and 95–104

Fixed. Also added an extension method on `IReporter` to verify that its verbosity is at least as specified, so checks become cleaner and less error-prone.

> ### Question — `TestSolutionAsync` `--output` direction

One more inverted check. Thanks for the catch.

Please review again.

rdeago:
> **`InvocationKind.Informational` is defined but never used** — `DotNetService.InvocationKind.cs`

It's used now.

Issue #269, opened 2026-05-23
Config Phase 2 — wire existing `bv` options to `buildvana.json`


rdeago:
Sub-issue of #267. Depends on Phase 1 (#268) and #280 (BuildPipeline extraction).

## Scope

- Make the `*Settings` injectables config-aware: feed `BuildvanaConfig` into `ReleaseSettings` and the new per-step Settings (`DotNetSettings` + `Restore/Build/Test/PackSettings`, hanging off the `BuildPipeline` from #280), so each `Resolve*` establishes the `flag → config → default` chain. `*Settings` stay the source-merging injectables; config DTOs stay in `Buildvana.Core.Configuration`.
- Wire: `dotnet.configuration` and `release.configuration` (collapsing the two hardcoded `"Release"` defaults; resolution `--configuration` → `release.configuration` → `dotnet.configuration` → `"Release"`, with `release.configuration` layering over `DotNetSettings.Configuration`); `release.checkPublicApi`; `release.changelogUpdates` (`none|stable|all`) + `release.emptyChangelog` (substitute text, else fail) — **config-only**, removing the `--unstable-changelog` and `--require-changelog` flags with no CLI replacement (changelog policy is repo-stable, not per-invocation); `release.dogfood`; `dotnet.*.args` and `restore.*.env` (where `*` is one of `restore`/`build`/`test`/`pack`/`nugetPush`, used according to the specific invocation in `DotNetService`) together with `dotnet.all.args` and `dotnet.all.env` (to be merged with invocation-specific settings by `DotNetService.RunDotNetAsync()`).
- `release.generateDocsFrom` (regex, short branch name) replaces `mainBranch`: delete `GitService.MainBranch`/`FindMainBranch`, the `--main-branch` global flag and its pre-parsing, and the discovery dance; the changelog permalink uses the actual `CurrentBranch`.

## Acceptance criteria

- Each option resolves flag-over-config-over-default; with no config file, behavior matches today (except the removed `--main-branch`).
- `changelogUpdates: none` skips changelog updates; an empty Unreleased section is substituted from `emptyChangelog` when present, else fails the release.
- Docs generation is gated by `generateDocsFrom` (default `["^main$", "^master$"]` when unconfigured, matching today's main/master discovery); `--main-branch` and `MainBranch` no longer exist.
- CHANGELOG: Unreleased entries for the new options; `**BREAKING CHANGE**` for the removed `--main-branch` flag.

PR #289, opened 2026-05-31
Honor value/element nullability in JSON schema; make child-process environment configurable

rdeago:
Preparatory work for #269 (Config Phase 2). No CLI surface changes; sets up the schema and process-runner plumbing that #269 will wire to `buildvana.json`.

### JSON schema value nullability
- Reconcile the nullability `System.Text.Json`'s exporter forces onto every reference-type dictionary value and collection element against the model's declared nullability (`string` vs `string?`), read from the owning member via `NullabilityInfoContext`. `null` is now kept only where the model actually declares it, recursing through nested generics.
- Allow `null` values for `dotnet` environment variables (`DotNetInvocationConfig.Env` → `IReadOnlyDictionary<string, string?>?`); the generated schema now accepts `["string","null"]` for those values.
- Remove the unused `JsonSchemaValidationException` and `JsonSchemaValidator.ValidateAndThrow`.

### Configurable child-process environment
- Thread an optional `environment` argument through `IProcessRunner.RunAsync` → `ProcessRunner` (CliWrap `WithEnvironmentVariables`, overlaying the inherited environment; a `null` value unsets a variable) → `DotNetService.RunDotNetAsync`. Not yet sourced from config — #269 connects it.

### Static analysis & tooling
- Move static-analysis configuration from `.globalconfig` into `.editorconfig`, with a test-specific overlay; refresh ReSharper settings; add `JetBrains.Annotations.Sources` to all C# projects; resolve/suppress the remaining ReSharper diagnostics; set JSON language mode for SARIF files in VS Code.

### Fixes
- `dotnet nuget push` has no `--verbosity` option; route `NuGetPushAllAsync` through an invocation kind that streams output without appending verbosity, so package pushes don't fail.

No CHANGELOG entry: nothing here changes released behavior (the push fix corrects a regression introduced earlier on this same branch), and the `env` config knob has no effect until #269 wires it.


rdeago:
This PR has been reviewed locally.

PR #310, opened 2026-08-01
Cover GitHub URL composition with tests, and make release-commit ordering a precondition

rdeago:
Follow-up to the two fixes pushed straight to `main` after the 2.1.2-preview release went out tagged as `2.1.3-preview`. Those stopped the bleeding; this closes the two holes that let it happen.

## Give GitHub URL composition an owner, and tests

The missing separator fixed in 6bba582 (`.../Buildvanareleases/tag/1.1.10`) was possible because nothing owned the contract: the repository URL was built in the adapter's constructor without a trailing slash, and two unrelated methods appended path segments to it. The two ends could drift, and did, for three releases.

`GitHubRepositoryUrls` now owns both ends — it takes the host, owner and name, builds the repository URL, and derives the release and file URLs from it. `GitHubServerAdapter` holds one and delegates. It needs no service provider, no Git origin and no token, which is precisely why the methods it replaces had no tests; there are now 22 cases covering both regressions, the enterprise-host case, normalization and the guards.

Two further bugs fell out of writing those tests:

- `Guard.IsTrue` takes the **parameter name** as its second argument, not a message, so both guards were reporting `Parameter "A path must be relative to be converted to a file URL." must be true, was false`. They now use the three-argument overload.
- The escape check only rejected a **leading** `..`, but `Uri` collapses parent segments as it parses, so enough of them anywhere in the path walked out of the repository and even out of the owner: `docs/../../../../../../etc/passwd` resolved to `https://github.com/etc/passwd`. Any `..` segment is now rejected, while a name merely *starting* with two dots is still accepted.

The second one is only reachable from Buildvana's own code, which passes a literal `CHANGELOG.md`, so it was never exploitable — but it is exactly the class of mistake the guard exists to catch.

## Require the release commit before a post-release commit

`AddPostReleaseCommit` called `EnsureReleaseCommit` on its caller's behalf, so it could move the Git height — and with it the version — in the middle of a call whose message the caller had already formatted from that same version. That is how the self-reference commit came to announce `2.1.2-preview` while the release it sat on was tagged `2.1.3-preview`.

`_version.Update()` is called in exactly one place in the codebase, inside `EnsureReleaseCommit`, so the version can only move at that single moment. Requiring the release commit up front removes the hazard outright rather than working around it with a deferred message factory: once the commit exists, nothing a caller invokes can change the version underneath it.

`UpdateRepository` keeps creating the commit implicitly, and the asymmetry is deliberate: it *amends* the release commit, so it owns it, and builds its message internally once the version has settled. A post-release commit merely sits on top of that commit, so it depends on it.

No functional change today — `ReleaseCommand` already settles the release commit before packing, which is what made that call legal in the first place.

## Notes

No changelog entry: every type involved is internal, and the user-visible URLs were already covered by the entry for 6bba582.

Verified with `dotnet bv pack` (314 tests, 0 warnings) and `inspectcode --severity=WARNING` (0 results).

### DIFF
--- a/src/Buildvana.Tool/Services/ServerAdapters/Internal/GitHub/GitHubRepositoryUrls.cs
+++ b/src/Buildvana.Tool/Services/ServerAdapters/Internal/GitHub/GitHubRepositoryUrls.cs
@@ -1,0 +1,84 @@
+﻿// Copyright (C) Tenacom and Contributors. Licensed under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using CommunityToolkit.Diagnostics;
+
+namespace Buildvana.Tool.Services.ServerAdapters.Internal.GitHub;
+
+/// <summary>
+/// Builds the URLs of a GitHub repository and of the resources it contains.
+/// </summary>
+/// <remarks>
+/// <para>This type owns both ends of the same contract: how the repository URL is built, and how the URLs of
+/// resources are derived from it. Keeping them together is the point — deriving a resource URL elsewhere,
+/// by appending to the repository URL, is what previously produced links with no separator between the
+/// repository name and the first path segment.</para>
+/// <para>The repository URL deliberately carries no trailing slash, so it can be displayed and compared as
+/// the canonical URL of the repository. Callers must therefore not treat it as a base URL for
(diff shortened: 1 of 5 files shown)


rdeago:

PR #319, opened 2026-08-03
Ship typed configuration and hook contexts in a new Buildvana.Runtime package

rdeago:
Follow-up to #312/#318, replacing the SDK-injected hook loader sources before they ever ship.

- **New packaged library `Buildvana.Runtime`**: the typed model of `buildvana.json`/`buildvana.jsonc` plus the run-time information `bv` shares with repository-owned hooks. Serialization is source-generated, so the same types work in file-based apps, where reflection-based JSON serialization is disabled. Hooks reference it with an unversioned `#:package Buildvana.Runtime` directive; Buildvana SDK pins the version to its own for every file-based app built in the repository.
- **Typed, per-hook contexts**: `PostReleaseHookContext` (shared verbatim between bv and hooks) structured into `Paths` and `Release` sections; context files move to `.buildvana-temp/hook-contexts/<command>/<moment>.json` so hooks can't clobber each other's replay context.
- `Buildvana.Core.Configuration` keeps schema validation and diagnostics on top of the packaged model; `nuget.feeds` is now a typed record (same JSON shape, better schema descriptions).
- Docs (`ReleaseHooks.md`), CHANGELOG, and architecture rules updated.

Note: `.github/workflows/claude.yml` was also updated in this branch (model/effort settings for the Claude workflow).


rdeago:

rdeago:
Good catch — confirmed and fixed in 83448fe, and the CPM branch turned out to have its own variant of the problem.

**Non-CPM:** a versioned `#:package Buildvana.Runtime@X` directive materializes as `<PackageReference Include="Buildvana.Runtime" Version="X" />` in the project body (verified via `dotnet project convert`), so the unconditional `Update` did clobber it. The `%(Version)` condition one-liner isn't legal at evaluation time (MSB4191), and the `Remove` idiom from `Sdk.targets` doesn't apply here (we're stamping metadata on the directive's item, not injecting a competing one), so the fix uses an item-function condition instead: the pin now only applies when the existing reference has no `Version` metadata.

**CPM:** a versioned directive under CPM fails upstream with NU1008 — the SDK generates `Version`, not `VersionOverride`, so the old comment's "use VersionOverride instead" advice was impossible to follow. The only override channel under CPM is the user's own `PackageVersion` item, which the previous unconditional `PackageVersion Include` would have duplicated (NU1506). The pin is now likewise guarded: a user-supplied `PackageVersion` wins.

All four scenarios (non-CPM ±explicit version, CPM ±user pin) verified against the edited targets file with a test harness.

On the two non-blocking notes, no changes in this PR:

- **`ReleaseCommand` patch coverage:** this is a pre-existing gap, not a regression — `ReleaseCommand` has no unit tests exercising hook-context assembly today, and building that harness is out of scope for this PR. It's a reasonable follow-up issue given how central `PostReleaseHookContext` now is to the hook contract.
- **Lockstep-release assumption:** deliberate and documented. `bv`, the SDK, and `Buildvana.Runtime` ship from this repo in lockstep by design; if that ever changes, the additive-only contract (and `UnmappedMemberHandling.Disallow`) would indeed need revisiting, but there are no plans in that direction.

@claude please review the PR again.

PR #321, opened 2026-08-03
Verify the pinned Buildvana SDK version before running SDK-using commands

### LINKED ISSUE #317: `bv` should verify the repo's pinned Buildvana SDK version before running commands that involve the SDK
### Reference version

2.1.41-preview

### Background and motivation

`bv`, `Buildvana.Sdk`, and `Buildvana.Runtime` are released in lockstep and designed as a matched group. The SDK forces the referenced version of `Buildvana.Runtime` for hooks, but nothing checks that the repo's pinned SDK version (`global.json`, `msbuild-sdks`) matches the running `bv`. A mismatch — a half-updated repo, a globally installed newer tool against an older pin — currently produces silent behavior drift instead of a clear message.

A typical situation where this can bite is a hook trying to load a Buildvana configuration, just validated by `bv` version X upon startup, using the loader from `Buildvana.Runtime` version Y, where X and Y have incompatible configuration shapes.

### Proposed enhancement

- `bv` subcommands declare, via a `usesSdk` boolean parameter in `ImplementsCommandAttribute` (default: `false`), whether they use the SDK. "Using the SDK" means, of course, invoking a `dotnet`, `msbuild`, or `nuget` command, or anything that executes any MSBuild target on any project in the solution. As a general rule, build pipeline commands (with the possible exception of `clean`) and `release` have `usesSdk = true`.
- Before invoking a subcommand that has `usesSdk = true`, `bv` reads the `Buildvana.Sdk` entry from `global.json` `msbuild-sdks` and compares it to its own version. A missing entry counts as a mismatch.
- On mismatch, `bv` fails with a message naming both versions and the fix for each direction (`dotnet tool update` vs. editing the pin).
- A new `sync-sdk` subcommand (name provisional) rewrites the pin to `bv`'s own version if the latter is more recent (a missing entry counts as less recent than any version); if the inverse is true (`bv` version predates the SDK version), the `tools.bv.version` property is rewritten in `.config/dotnet-tools.json` _if present and if equal to the currently running `bv` version_ (a `dnx bv` invocation should not touch `dotnet-tools.json`). If the two versions are equal, the command modifies neither. If `global.json` is missing entirely, it is created with only an `msbuild-sdks` section.

Acceptance criteria:

- Check runs before every subcommand that uses Buildvana SDK.
- Clear failure message naming both versions and mentioning the `bv sync-sdk` command.
- Tests for match, mismatch (both directions), and missing-pin handling.
- Changelog entry under `## Unreleased changes`.

### Non-goals

Edge cases triggered by the user by pinning versions in the wrong places (e.g., a version-suffixed `Sdk` attribute in a project file, or a pinned `Buildvana.Sdk` in either `Directory.Build.props` or `Directory.Build.targets`) are not a concern. Developers usually know what they are doing; when they don't, let them have a learning experience.

### Implementation proposals

- A `SelfVersionService` (name provisional) doing checks and JSON edits via the existing JSON helpers (for `global.json`) or `dotnet tool` (for `.config/dotnet-tools.json`).
- Versions are compared according to SemVer, i.e., preview < stable. A missing pin counts as less recent than _any_ version. SemVer metadata don't participate in comparisons and are never written to pins.
- A global `--skip-sdk-check` option disables the check, for scenarios that require a deliberate mismatch (e.g., bisecting an SDK regression in CI).

### Risks

None in sight.

---

🤖 Edited with [Claude Code](https://claude.com/claude-code)

rdeago:
Closes #317.

`bv`, `Buildvana.Sdk`, and `Buildvana.Runtime` are released in lockstep and designed as a matched group, but nothing checked that the repo's pinned SDK version matches the running `bv`. This PR adds that check, plus the command that fixes a mismatch.

### What's here

- Subcommands declare SDK usage via a new `usesSdk` parameter of `ImplementsCommandAttribute`; `restore`, `build`, `test`, `pack`, and `release` declare it. Before running such a command, `bv` compares the `Buildvana.Sdk` entry under `msbuild-sdks` in `global.json` with its own version (SemVer precedence, build metadata ignored). A missing file, section, or entry — or an unparseable pin — counts as a mismatch. The failure message names both versions and every way out. The new global option `--skip-sdk-check` skips the check, for deliberate-mismatch scenarios such as bisecting an SDK regression in CI.
- The new `bv sync-sdk` command aligns the two, updating whichever is older: it rewrites (or creates) the `global.json` pin when the pin side is older, and runs `dotnet tool update bv` when the tool side is older — the latter only when the running `bv` comes from the repository's tool manifest, so e.g. a `dnx bv` invocation never touches `.config/dotnet-tools.json`.
- Supporting work: `IJsonHelper.InsertProperty`, a byte-preserving JSON property insertion (formatting, comments, CRLF/LF, and BOM survive) used to add missing pins; a `FakeProcessRunner` capture fake in `Buildvana.Core.Testing`; a new `Buildvana.Core.Json.Tests` project.

### Bug fixed in passing: `bv clean` vs. hook build caches

The first commit replaces `dotnet clean <hook>.cs` with direct deletion of each hook's runfile artifacts directory (computed with the same scheme as the SDK's `VirtualProjectBuilder`, guarded by an SDK-conformance test). `dotnet clean` on a file-based app requires a restorable SDK and leaves both build-cache markers in place, so it never actually reset the app — worse than useless. No separate issue was filed: the hooks feature shipped hours before this fix with zero users. This fix is also what lets `clean` stay `usesSdk = false`, resolving the "possible exception of `clean`" left open in #317: cache wiping no longer evaluates any project, and evaluation was the only SDK dependency `clean` had.

### Verification

- 452 tests pass solution-wide, including exact-bytes assertions on every `sync-sdk` write path and message-content assertions on every check failure.
- `dotnet bv pack` green; ReSharper `inspectcode --swea --severity=WARNING`: zero results.


rdeago:

rdeago:
Thanks for the review. All four accepted points landed in 68f1ffa, with one correction to the suggested fix for the cache-root bug.

**Docs (finding 1):** `docs/DirectoryStructure.md` now prescribes the `global.json` pin scheme: versionless `Sdk.props`/`Sdk.targets` imports, plus a new `global.json` section covering `msbuild-sdks`, the SDK version check, `--skip-sdk-check`, and `bv sync-sdk`. The `Version`-attribute form is gone from the docs — the check, `sync-sdk`, and self-reference updating all assume the `global.json` scheme, so it is no longer a documented configuration. BVSDK1002 is reframed as the safety net for stray `Version` attributes. The tree diagram gains `global.json` and `.config/dotnet-tools.json`.

**Cache root (finding 2):** the bug was real, but the suggested fix mischaracterized the SDK: `VirtualProjectBuilder.GetTempSubdirectory` does _not_ pass `SpecialFolderOption.DoNotVerify` — it makes the same `GetFolderPath` call bv does and **throws** (`EmptyTempPath`) when the result is empty. Since the SDK therefore fails every file-based-app build in that state, no cache can exist there, so `GetArtifactsDirectory` now returns `null` for "this machine has no cache root" and `bv clean` skips with an info message instead of resolving a relative path against the working directory. Your `Path.IsPathRooted` test suggestion is in.

**`sync-sdk` half-synced state:** kept the never-touch-the-manifest-implicitly stance, made it legible: after writing the pin (and on the already-in-sync path too, so repeated runs stay honest), `sync-sdk` warns when the tool manifest still pins a different bv, naming both versions and both ways to finish the alignment. Covered by three new tests via a new `CaptureReporter` fake in `Buildvana.Core.Testing`.

**`InsertProperty` minors:** doc-only — the `parentPath` docs now state that array elements contribute no path segment (so a path can lead into an array element's descendants) and that callers must not pass array-traversing paths; the remarks admit the two-space fallback indent when expanding an empty object. The ungated child output stays as is: `DotNetService` streams file-based-app output with the same `null` gate, and the `dotnet tool update` output is the command's deliverable. The shared-constants refactor, `NuGetVersion.Parse` hardening, and message-precision nits were deliberately skipped.

Full gate is green: `dotnet bv pack` (455 tests passing, 0 warnings, 0 errors) and ReSharper inspectcode at WARNING severity reports zero findings.

@claude please review the new changes.

rdeago:
Thanks for round 2. All five landed, one commit per finding (fa13bdf..a739f1b).

**Advisory read (finding 1, fa13bdf):** `WarnIfToolManifestDisagrees` now catches `BuildFailedException` around its manifest read and degrades to a Detail message, so a corrupt `.config/dotnet-tools.json` can no longer fail a sync-sdk that has already written the pin; `UpdateToolManifestAsync`'s load-bearing read still throws. New test: malformed manifest → sync succeeds, pin written, no warning, one Detail message naming the manifest.

**Direction-aware remedy (finding 2, 1f5def7):** the warning now branches on `VersionComparer.VersionRelease.Compare`: an older manifest keeps the two-remedy advice (`dotnet tool update bv --version <own>`, or re-run sync-sdk through the manifest), a newer manifest advises `dotnet bv sync-sdk` so the manifest's bv re-pins the SDK to its own version — the downgrade advice is gone from that direction. The diagnosis names both versions in both directions. A new test covers the newer direction (asserting the `dotnet tool update` advice is absent), and the older-direction test now pins its remedy too.

**CaptureReporter (finding 3, 5d5746e):** took the drop-the-`init` option: `Verbosity` is now get-only, fixed at `Diagnostic`, and the class doc states the guarantee — including why it matters (gating helpers consult it before calling `Report`).

**Tool manifest docs (finding 4, edbb43e):** `.config/dotnet-tools.json` got its own section in `DirectoryStructure.md` (with TOC entry), covering what it pins, that it's optional (global installs and `dnx` exist), and its two roles in sync-sdk — provenance guard and disagreement warning; cross-linked both ways with the `global.json` section.

**BVSDK1002 (finding 5, a739f1b):** the table row now describes the actual condition — imports from two different versions of Buildvana SDK — and points at stray `Version` attributes as the usual cause. Kept terse: MD060 enforces aligned pipes in that table, and the full global.json-pin context lives one link away in `DirectoryStructure.md`.

The two trivia stay as they are: the double `File.Exists` buys the better no-manifest error message, and hoisting the null check out of `CleanBuildCaches`'s loop would require a separate cache-root API on `FileBasedAppHelper` just for cosmetics.

Full gate is green: `dotnet bv pack` (457 tests passing, 0 warnings, 0 errors) and ReSharper inspectcode at WARNING severity reports zero findings.

@claude please review the new changes.

rdeago:
Thanks for round 3. Both fixable findings and the docs note landed, one commit per finding (620e41a..94a08ad).

**Denied-access errors (finding 1, 620e41a):** fixed in `JsonHelper` as suggested, with two widenings. The catch filter matches the tier's existing idiom in `VersionFile` (`IOException or UnauthorizedAccessException or SecurityException`) rather than adding only `UnauthorizedAccessException`; and `InsertProperty` got the same treatment — it wasn't in your list, but it shares the read/write pattern and the same `IJsonHelper` contract, so leaving it out would have kept the contract violated in one method. Four new tests (one per method) pass a directory as the path — which raises `UnauthorizedAccessException` on every platform — and assert the `BuildFailedException` wrap.

**Warning level (finding 2, 0134841):** accepted your revised take: the advisory degrade now reports at Warning instead of Detail — visible at default verbosity, no effect on the exit code. The comment above the catch records why. The malformed-manifest test was renamed and now asserts one warning naming the manifest and zero Detail messages.

**Provenance wording (finding 3, 94a08ad):** the tool-manifest bullet now states the guard as what it is — the running bv's version matching the manifest's `bv` entry — explains that version equality is how sync-sdk decides the running bv is the manifest's, and notes the coinciding-version case is harmless: the manifest ends up pinning exactly the version `global.json` asks for. The `global.json` section's one-line summary of the same guard was aligned too, which also swallowed your second trivia item — the forward-slash manifest path sat on the same line and now follows the backslash convention.

The transient-write trivia stays as is, per your own assessment: not a regression, correct end state, and the warning discloses which version wins.

Full gate is green: `dotnet bv pack` (461 tests passing, 0 warnings, 0 errors) and ReSharper inspectcode at WARNING severity reports zero findings.

@claude please review the new changes.

rdeago:
Round 4 addressed in `7dfadb6..7498db5`, one commit per finding.

**1. `CreateGlobalJson` write wrapped (7dfadb6).** On the caveat you raised: we decided it is _not_ a deliberate tier-wide stance — the other unwrapped sites are the same latent gap, just out of this PR's scope. This PR hardens the `sync-sdk` path, so its one remaining raw-stack-trace door is closed here with the exact `VersionFile`/`JsonHelper` filter, plus a test that routes `SyncSdkAsync` into the create-new-file path via a directory named `global.json` (same cross-platform trick as `DeniedAccessPath`, asserting the inner `UnauthorizedAccessException`). The rest of the sites — including unwrapped _reads_ your list didn't cover — are now #324.

**2. Inner exception pinned (e3abc92).** Each of the four `*_WithDeniedAccess_Fails` tests now asserts `InnerException` is `UnauthorizedAccessException`, so a temp-path environment quirk degrades to a loud failure instead of vacuous coverage.

**3. Re-run wording fixed (7498db5).** We took the Angular CLI's message as the model (state which binary runs, don't assume the caller knows): the post-update info now reads _"bv updated to X in the tool manifest. Re-run your command with 'dotnet bv', which always runs the manifest's version."_ — which gives correct advice in the coinciding-version corner case without having to call it out. The deeper ng lesson (a non-local bv delegating to the manifest-pinned version outright) is being drafted as a separate enhancement proposal.

Local gate is green: `dotnet bv pack` (build + tests + artifacts) with zero warnings, ReSharper `inspectcode --swea --severity=WARNING` with zero results.

rdeago:
Round 5 addressed in `63e705d..1f33925`, one commit per item — trivia included, deliberately: every trivium this series has left standing has come back as a finding the next round, so this round leaves nothing behind.

**1. Provenance wording swept to the last two copies (63e705d).** The changelog bullet now states the guard as version equality against the manifest's `bv` entry — coinciding-version escape included, harmless as the docs already note — and gained a sentence on the tool-manifest disagreement warning, which was indeed missing from the release-notes copy. The `SyncSdkAsync` remarks got the same correction, so the contract `UpdateToolManifestAsync`'s comment defers to now matches what the code checks. A repo-wide grep for the absolute claim and its variants confirms these were the only two copies left.

**2. Post-update info pinned (54cf719).** `SyncSdkAsync_WithNewerPin_UpdatesToolViaDotnetToolUpdate` now passes a `CaptureReporter` and asserts the last info names the target version and routes the re-run through `'dotnet bv'`, via a new `InfosOf` helper next to `WarningsOf`/`DetailsOf`.

**3. `<exception>` tag reworded as a headline (1f33925).** Took the headline option: after 7dfadb6 the `BuildFailedException` doors are too many for a list that stays honest, so the tag now says sync failures surface as `BuildFailedException` with the reason in the message, and stops reading as exhaustive.

Full gate is green: `dotnet bv pack` (462 tests passing, 0 warnings, 0 errors) and ReSharper inspectcode at WARNING severity reports zero findings.

@claude please review the new changes.

rdeago:
Round 6 addressed in `8d00f01..a53bb6c`, one commit per finding.

**1. Three outcomes stated, third clause tested (8d00f01).** The remarks now say what the code does: a pin equal to bv's version ignoring build metadata changes nothing; an older, missing, or unparseable pin is rewritten; a newer pin updates bv. `SyncSdkAsync_WhenInSync_ChangesNothing` is parameterized like its `EnsureSdkVersionMatch` sibling — exact match plus both build-metadata rows — so the early return is pinned against exactly the case the old sentence got wrong.

**2. Sibling `<exception>` tag headlined (a53bb6c).** `EnsureSdkVersionMatch`'s tag got the same treatment as `SyncSdkAsync`'s: the check failed and the message names the reason — the unreadable-`global.json` door is covered by not enumerating at all.

Your calls on the trivia stand as written — all three stay as they are, for the reasons you gave.

Full gate is green: `dotnet bv pack` (464 tests passing, 0 warnings, 0 errors) and ReSharper inspectcode at WARNING severity reports zero findings.

@claude please take another look — and from this round on, flag only merge-blocking issues; anything else, list briefly as non-blocking notes that need no re-review.

PR #326, opened 2026-08-05
Reject unknown options on commands that take none

### LINKED ISSUE #314: Commands without their own options silently ignore unknown options
### Affected version(s)

2.1.12-preview (and current `main`)

### Is this a regression?

NO - This bug also affected the previous version

### Context

Option validation for `bv` subcommands lives in two places: `CommandArgumentValidator` (rejects forwarded tokens and excess positionals) and each command's `*Settings.Parse`, which throws on `reader.Remaining.Count > 0`. Commands with no settings type (`clean`, `version show`) have no `Parse` — so their option tokens are never inspected by anyone.

### Expected behavior

```
> bv clean --bogus
error: Unknown option '--bogus' for command 'clean'.
```

### Actual behavior

`--bogus` is silently ignored and the command runs normally. Typos in real options or options meant for another subcommand go unnoticed.

### How to reproduce

`dotnet bv clean --bogus` — exits 0, cleans as if nothing happened.

### Configuration

`bv` CLI tool, any OS.

### Other information

Likely fix: in `CommandArgumentValidator`, reject any leftover option token for non-forwarding commands whose `SettingsType` is `null` (commands _with_ settings keep their existing `Remaining` check).

Acceptance criteria:

- unknown options are rejected for settings-less commands;
- tests cover both settings-less and settings-carrying commands;
- CHANGELOG entry under _Unreleased changes_.

rdeago:
Closes #314.

Option validation lived in two places: `CommandArgumentValidator` (forwarded tokens, excess positionals) and each command's `*Settings.Parse` (leftover option tokens). Commands with no settings type — `clean`, `sync-sdk`, `version show` — fell through both nets, so `bv clean --bogus` ran normally and typos went unnoticed.

`CommandArgumentValidator` is now the single authority for unknown-option rejection: it consumes every option declared via `BvOptionAttribute` on the command's settings type (`ValueName` tells it which options swallow the following token) and rejects whatever is left, as `Unknown option '--bogus' for command 'clean'`. A settings-less command is simply one that declares zero options — no special case. The now-redundant leftover checks in `ReleaseSettings.Parse` and `VersionAdvanceSettings.Parse` are gone, so a future `*Settings` type cannot reintroduce the bug by forgetting the check. Offending tokens are reported in command-line order (excess positionals before unknown options), and — since validation runs before the SDK version check — `bv release --bogus` now reports the typo even in a repository with a mismatched SDK pin.

Note: the issue names `clean` and `version show`; `sync-sdk` is settings-less too and is covered by the same fix.

Note: the first review round suggested centralizing unknown-option rejection as a follow-up issue. It is folded into this PR instead: the changes are small and self-contained, and a separate PR wasn't worth the overhead.

### Verification

- New/updated tests: unknown options rejected for settings-less commands (`clean --bogus`, and `version --bogus` pinning the canonical `version show` name in the message) and for settings-carrying commands (`release --bogus`); declared options consumed correctly (space-separated and inline values, flags, missing-value error); `clean junk --bogus` reports `junk` first.
- 469 tests pass solution-wide; `dotnet bv pack` green; ReSharper `inspectcode --swea --severity=WARNING`: zero results.


rdeago:

rdeago:
All four findings addressed, one commit per finding:

- **Nit 1** (c28a18a): excess positionals are now checked before unknown options, so `bv clean junk --bogus` reports `junk`; a new test pins it. One accepted limitation: `bv --bogus clean junk` also reports `junk` first — true left-to-right ordering would require tracking original token positions, which isn't worth it for that shape.
- **Nit 2** (bba4f19): both settings-less tests now assert the exact message; the alias test pins the canonical `version show` name.
- **Follow-up** (2babf9c): folded into this PR rather than spun off — the changes are small and self-contained. `CommandArgumentValidator` is now the single authority: it consumes every `BvOptionAttribute`-declared option (`ValueName` marks the value-swallowing ones) and rejects the remainder; a settings-less command is just the zero-options case. The redundant `Remaining` checks in `ReleaseSettings.Parse` and `VersionAdvanceSettings.Parse` are gone — plus the equally redundant excess-positional check in `VersionAdvanceSettings.Parse` — and the timing asymmetry is fixed: `bv release --bogus` reports the typo before the SDK version check. `BvOptionAttribute`'s "help metadata only" remark was updated to match its new validation role.
- **Editorial** (6653a91): the changelog entry is reframed around `clean` plus the general rule, as suggested, and now also mentions the report-before-SDK-check timing.

Verification: 469 tests pass solution-wide, `dotnet bv pack` green, ReSharper `inspectcode --swea --severity=WARNING` zero results.

@claude please take another look. From this round on, flag only merge-blocking issues; anything else, list briefly as non-blocking notes that need no re-review.

PR #327, opened 2026-08-05
Route `bv` narration to standard error, keeping standard output for deliverables

### LINKED ISSUE #315: Route `IReporter` diagnostics to standard error, keeping standard output for deliverables
### Reference version

Current `main` + #313

### Background and motivation

#313 introduced the results-vs-logs distinction: `bv version show` prints its report unconditionally, while diagnostics flow through `IReporter` gated by verbosity. But `ConsoleReporter` writes everything except child-process stderr to **stdout**, so at `--verbosity normal` a piped `bv version show | …` still receives `info:` lines mixed into the result stream.

The universal CLI convention (git, npm, curl, cargo, kubectl, gh; also the default log destination in Python, Go, Rust, and Java runtimes) is: results → stdout, diagnostics/progress → stderr, so results stay pipeable at any verbosity. The notable counterexample is MSBuild/dotnet, whose build log goes to stdout because there the log _is_ the deliverable — which this proposal respects (see below).

### Proposed enhancement

All leveled reporter messages (`error:`/`warning:`/`info:`/`detail:`/`trace:`) and activity header/outcome lines go to stderr. Command deliverables (e.g. `version show`'s report) and `ChildOutput` (child `dotnet` stdout, which _is_ the payload of build-ish commands) stay on stdout. `ChildError` already goes to stderr.

### Implementation proposals

Change `ConsoleReporter` write targets. Color detection must then probe `Console.IsErrorRedirected` for reporter output (currently only `IsOutputRedirected` is consulted).

### Usage examples

- `bv version show | some-parser` works at any verbosity;
- `bv build 2>build.log` separates bv's narration from `dotnet`'s output.

### Risks

Behavior change for scripts/CI that capture bv's stdout expecting diagnostics in it — worth a **BREAKING CHANGE** changelog entry under _Changes to existing features_. Interleaving of stdout and stderr in terminals/CI captures can reorder lines. Mitigated by the fact that per-command default verbosity (#313) already minimizes diagnostics where piping matters most.

### Additional information

Filed as follow-up from the #313 discussion.

Acceptance criteria:

- reporter output on stderr, deliverables and `ChildOutput` on stdout;
- color handling correct for both streams;
- **BREAKING CHANGE** CHANGELOG entry under _Changes to existing features_.

rdeago:
Closes #315.

## What changed

`bv` now follows the prevailing CLI convention (git, npm, cargo, kubectl, …): results go to standard output, narration goes to standard error, so results stay pipeable at any verbosity.

- `ConsoleReporter` writes all leveled diagnostics (`error:`/`warning:`/`info:`/`detail:`/`trace:`) and activity header/outcome lines to standard error. `ChildOutput` (child `dotnet` standard output — the payload of build commands) stays on standard output; `ChildError` was already on standard error.
- The startup logo also moves to standard error: technically a separate concern, but piped output isn't clean until it moves too. `bv --version` output and help remain standard-output deliverables.
- Color auto-detection now probes `Console.IsErrorRedirected` (plus `NO_COLOR` and, on Windows, whether virtual-terminal processing can be enabled on the standard error handle). `--color`/`--no-color` still override everything.

## Design notes

- Labels are colored with raw ANSI escape sequences (new `AnsiEscapes` class) instead of `Console.ForegroundColor`. The BCL implements `ForegroundColor` against standard output — on Unix it emits the escape sequences _to standard output_ — so with diagnostics on standard error it would corrupt a redirected deliverable stream (e.g. `bv build > log`). `AnsiEscapes.Foreground(ConsoleColor)` maps all 16 `ConsoleColor`s to the SGR codes the BCL uses on Unix, so `error:`/`warning:` render exactly as before.
- New `VirtualTerminal.TryEnableOnStandardError()` enables `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on the Windows standard error handle (ConPTY hosts like Windows Terminal already interpret VT; legacy conhost needs the explicit opt-in). On failure, auto-detected color turns off. Returns `true` outright on non-Windows.
- `ConsoleReporter` finally has unit tests (new `Buildvana.Core.ConsoleOutput.Tests` project, 43 tests): per-level stream routing, verbosity gating, label-only coloring, activity lines, child passthrough.

## Breaking change

Scripts and CI steps that captured diagnostics from `bv`'s standard output must now capture standard error (e.g. via `2>&1`). Changelog entry added under _Changes to existing features_.

## Coverage follow-up

Codecov flagged the patch coverage of the first push. Rather than just chase the number, this PR also establishes the project's coverage policy, now codified in `.claude/rules/testing.md`: testable code gets tested; code whose behavior is owned by the environment is excluded in source via `[ExcludeFromCodeCoverage(Justification = ...)]`, never through Codecov configuration.

Applied here:

- `AnsiEscapes.Foreground` now computes the SGR code from the `ConsoleColor` bit layout (the same red/blue bit swap the BCL uses on Unix) instead of a 16-arm switch; new `AnsiEscapesTests` hold the literal 16-color table as an independent oracle, plus boundary out-of-range throws.
- `--verbosity` parsing moved from `Program` to a new `VerbosityParser` in the CommandLine layer, with tests pinning the accepted aliases and the error message.
- New `ConsoleReporter` tests cover the unknown-`MessageLevel` throw and the double-dispose idempotence of activity scopes.
- Excluded with justification: `VirtualTerminal` (P/Invoke over console state), `Program` (composition root), `ConsoleReporter.DetectColor`/`IsNoColorSet` (process-global console/environment probes).


rdeago:

rdeago:
All findings addressed in five commits, one nit declined with rationale, and the optional refactor now has a home. Item by item:

**1. `--color` never enables VT on Windows** — fixed in [`27af538`](https://github.com/Tenacom/Buildvana/pull/327/commits/27af538c3349ab0abb1cb28f1e95f0fb7f21c8bb). The constructor now attempts `VirtualTerminal.TryEnableOnStandardError()` whenever the override forces color on; auto-detection keeps folding the attempt into `DetectColor`. Judgment call taken as you leaned: the explicit override stays authoritative even when enabling fails — the user asked.

**2. Non-Windows assumes any terminal understands VT** — fixed in [`230aaf3`](https://github.com/Tenacom/Buildvana/pull/327/commits/230aaf397d2dfef9eb83ed243f07d593767a6b9c). The non-Windows branch now requires a usable `TERM` (set, non-empty, not `dumb`) as a cheap proxy for the terminfo lookup `Console.ForegroundColor` used to do. It lives in `VirtualTerminal` as you suggested, since a `TERM` check in `DetectColor` would misfire on Windows.

**3. A leveled line is up to five auto-flushed writes** — fixed in [`4eec85b`](https://github.com/Tenacom/Buildvana/pull/327/commits/4eec85be8c1a8e9c3110c0c91625285727e8d9e5). `WriteLeveledLine` composes the whole line and issues a single `WriteLine`; output is byte-identical and the existing tests pass unchanged.

**Tests and coverage** — closed in [`077b9ec`](https://github.com/Tenacom/Buildvana/pull/327/commits/077b9ecf99d6013bec54dc0473628d8df6fd30c0):

- Out-of-order disposal is pinned through observable behavior: after disposing Outer while Inner is still open, a next sibling activity still nests at depth 2 (`[2] Next: starting...`), which blind popping would break; Outer's outcome line is asserted too.
- `ChildOutput` and `ChildError` each gained the satisfied-minimum case, and `ChildError_NoMinimumVerbosity_IgnoresVerbosity` completes the symmetry with its existing `ChildOutput` twin.
- `VerbosityParser.Parse(null)` throwing `ArgumentNullException` is pinned as part of the contract.

**Nits** — [`8ae30c0`](https://github.com/Tenacom/Buildvana/pull/327/commits/8ae30c020f115577e20496957022ad872b377df2):

- `ArgumentNullException` is now documented on `Parse`.
- The `AnsiEscapes` remarks claim color-index parity with `Console.ForegroundColor` rather than visual identity, noting the Unix BCL may emit terminfo-derived sequences instead of these literal codes.
- The `VirtualTerminal` remarks state that the mode change outlives the process and is deliberately not restored, per common CLI practice.
- **Declined:** dropping `SetLastError = true`. It stays on purpose — it documents the intended semantics of the imports and keeps error codes available should console misbehavior ever need diagnosing in the field. A comment above the imports now carries that rationale, so it no longer reads as leftover noise.

**`TextWriter` injection** — agreed on both counts: it is the honest design per our own testing rule, and it is not for this PR. Opened #328 so the follow-up actually exists instead of being politely mentioned by everyone and done by no one.

State after the five commits: 530 tests passing, zero warnings, ReSharper solution-wide inspection clean.

rdeago:
**Follow-on from fix 2** — agreed on both the diagnosis and which option the coverage rule prefers: the justification described a branch that no longer existed, and the `TERM` check is exactly the "pure logic extracted out of plumbing" case the policy asks for.

6770b62efd2a8e38f22604bdb0336a824fda17d9 extracts the decision as `VirtualTerminal.IsNonWindowsTerminalCapable(string?)` (internal, exposed to the test project via `InternalsVisibleTo`) and tests it directly, including pinning the Ordinal comparison as intentional (`"Dumb"` counts as capable). The class-level exclusion is gone; what remains excluded is scoped to what is genuinely environment-owned:

- `TryEnableOnStandardError` — reads `TERM` off Windows, manipulates the attached console on Windows;
- the three P/Invoke imports — their attributes merge into the generator-emitted implementations, so the marshalling stubs (never executed off Windows) stay out of the coverage denominator rather than surfacing as uncovered lines once the class-level attribute is removed.

State: 535 tests passing, zero warnings, ReSharper inspection clean.

PR #329, opened 2026-08-06
Fail cleanly on denied or failed user-file access

### LINKED ISSUE #324: File I/O outside IJsonHelper surfaces raw stack traces instead of clean errors
### Affected version(s)

2.1.41-preview

### Is this a regression?

NO - This bug also affected the previous version

### Context

#317 / PR #321 established the tier idiom for user-file I/O failures: catch `IOException or UnauthorizedAccessException or SecurityException` and rethrow as `BuildFailedException` with a `Could not read from/write to <path>: <reason>` message (see `VersionFile.Save` and `JsonHelper`). Several `bv` services still do bare `File.*` I/O that predates the idiom.

### Expected behavior

A denied or failed file access during any `bv` command produces a single clean error line naming the operation, the path, and the OS reason, with the `BuildFailedException` exit code.

### Actual behavior

The raw `UnauthorizedAccessException`/`IOException` escapes `Program.Main`'s handler and the user gets an unhandled-exception stack trace pointing at bv internals.

### How to reproduce

Make the target file read-only (or lock it in another process) and run the corresponding command, e.g. `attrib +r CHANGELOG.md` (or `chmod a-w`) before a changelog-writing operation.

### Configuration

All project types, all OSes; not configuration-specific.

### Other information

Known sites — writes: `ChangelogService.cs:228`, `ChangelogService.cs:292`, `SelfReferenceUpdater.cs:162`, `PublicApiFilesService.cs:136-137`, `HookRunner.cs:72`; reads: `PublicApiFilesService.cs:91`, `PublicApiFilesService.cs:111`, `PublicApiFilesService.cs:118`, `GitHubServerAdapter.cs:224` (`File.OpenRead` for release asset upload).

`WriteThisAssemblyConstantsFile.cs:61` in `Buildvana.Sdk.Tasks` has the same shape but lives in the MSBuild task layer (failure presents as MSB4018), so whether it's in scope is a separate call.

With this many call sites, consider a small shared helper in the Core tier instead of repeating the filter.

Acceptance criteria: all listed sites wrapped (or explicitly excluded with a reason), denied-access tests via the directory-as-target trick, and a `CHANGELOG.md` entry under Unreleased changes.

rdeago:
Closes #324.

A denied or failed file or directory access during a `bv` command used to escape `Program.Main`'s handler and surface as an unhandled-exception stack trace pointing at bv internals. File and directory accesses now report failure as a single clean error line naming the operation, the path, and the OS reason, and exit with the `BuildFailedException` exit code.

### Changes

- **New `Buildvana.Core.IO` project**: `UserFile` (read/write/append × text/lines/bytes, sync and async, encoding overloads, plus `Exists` and `OpenRead`/`OpenText`) and `UserDirectory` (`CreateDirectory`, `Delete`, `DeleteIfExists`, `Exists`, glob-based `EnumerateFiles`), wrapping environment-driven failures in `BuildFailedException` with the `Could not read from/write to {path}: {reason}` messages established by #321. Open methods guard the open only, as documented.
- **All known bare sites migrated**: the sites listed in #324 plus additional ones found while sweeping — `ChangelogService`'s and `SelfReferenceUpdater`'s reader opens, `ReleaseCommand`'s asset-list reads, `DocFxService`'s metadata write, `GitHubServerAdapter`'s `GITHUB_OUTPUT` append, `HookRunner`'s context-directory creation, the `bv clean` deletions and glob walks previously in the tool's `FileSystemHelper` (now dissolved into `UserDirectory`), and solution-file discovery in `HomeDirectorySolutionContextFactory` (on every command's path).
- **Narrow filters fixed**: `BuildvanaConfigLoader` and the two `Buildvana.Runtime` loaders caught only `IOException`, letting `UnauthorizedAccessException` escape; Runtime keeps its inline filter (BCL-only closure) but widened to the tier idiom.
- **Exception classification consolidated**: `Buildvana.Sdk.Internal.ExceptionExtensions` moved to `Buildvana.Core.Diagnostics` in `Buildvana.Core.Abstractions`, reshaped as C# 14 extension properties (`e.IsIORelatedException`, `e.IsFatalException`, …). User-file I/O thereby adopts the task layer's broader filter: the malformed-path family (`ArgumentException`, `NotSupportedException`) now produces clean errors too, while null/empty-path programmer errors still throw raw (guards sit outside the catch).
- **`Sdk.Tasks` keeps coded diagnostics**: `WriteThisAssemblyConstantsFile` wraps I/O failures inline in the `ConvertPfxToSnk` shape — `BVSDK1051` for writes, new `BVSDK1052` for the read-and-compare — because SDK error codes are a documented contract (`docs/SdkDiagnostics.md`). Routing them through `UserFile`, as an earlier revision did, dropped the codes.
- **Deliberate packaging trade-off**: dissolving `FileSystemHelper` moved the glob walk and its `Microsoft.Extensions.FileSystemGlobbing` reference into `Buildvana.Core.IO`, so the SDK package now carries `Microsoft.Extensions.FileSystemGlobbing.dll` (~47 KB, never loaded by tasks) in `bin\`. Accepted: a dedicated `Buildvana.Core.IO.Globbing` project would be exactly the pre-created split the Core tier layout avoids, and the package already ships multi-megabyte native binaries.
- **Excluded with reason**: other `Sdk.Tasks` sites (task layer, present as task errors via `IsFatalException` handling), `Core.Testing` temp-file helpers (test infra), `HeadTailPipeTarget` (process pipe, not a user file), and `SolutionPersistence`'s `OpenAsync` solution-file read (third-party I/O boundary, like Octokit and LibGit2Sharp). `ChangelogService`'s rewrites are guarded at open, but a mid-read failure window between validation and rewrite remains; closing it, with test coverage, is tracked in #330.

### Testing

- `Buildvana.Core.IO.Tests` suite (82 tests): denied-access via the directory-as-target trick for every `UserFile` member, the `UserDirectory` surface (delete/recursive/missing/reporter paths, glob enumeration, missing-base-directory, null-path raw throws), malformed-path coverage (embedded-NUL path, platform-stable), and round-trip happy paths.
- New `Buildvana.Core.Diagnostics.Tests` suite covering the four exception classifiers, including nested and `AggregateException` cases.
- `Buildvana.Runtime.Tests`: read-failure branch of both loaders via exclusively-locked files (portable: .NET maps `FileShare` to advisory locks on Unix).
- `HookRunner` consumer-level denied-access test (directory planted at the context-file path).
- Full solution: build and `inspectcode` clean at zero warnings; all test suites green; `bv pack` produces all three packages.

### Changelog

Entry added under _Unreleased changes → Bugs fixed in this release_.


rdeago:

rdeago:
All items addressed in 308bf2f.

1. **Gone further than suggested**: `FileSystemHelper` is dissolved into `Buildvana.Core.IO` entirely. `UserDirectory` gains `Delete(path)`/`Delete(path, recursive)` (wrapped BCL mirrors), `DeleteIfExists(path, reporter?)` (the recursive skip-if-missing convenience `bv clean` and `HookRunner` use), `Exists`, and `EnumerateFiles` — the glob walk moved along with its FileSystemGlobbing dependency, is now guarded, and returns a materialized list so enumeration failures report at call time. `UserFile` gains `Exists`. All call sites rewired.
2. Coded diagnostics restored in `WriteThisAssemblyConstantsFile`, inline in the `ConvertPfxToSnk` shape: `BVSDK1051` for writes, and since its text was indeed already wrong for reads, a new `BVSDK1052` ("could not be read") for the read-and-compare path, documented in `SdkDiagnostics.md`.
3. Runtime now has an `internal` `ExceptionExtensions.IsIORelatedException` of its own, hand-synced with the Core property named in its doc as the source of truth — same call-site idiom, still BCL-only.
4. Changelog reworded ("File and directory accesses now report…") and the `bv clean` example added. The `ChangelogService` streaming reads stay: early exit is the point of `OpenText` there.
5. Fixed with the validate-then-delegate split; a bad path now throws before the state machine starts. Pinned by tests that invoke the async members without `await`.
6. The reachable case is fixed: `SaveObject` serializes to memory first, so only the file write sits in the guarded region. The lazy-enumeration case stays latent and is now documented in `<remarks>` on all `*AllLines` members.
7. Null/empty raw-throw tests added (sync + async), `OpenText` no-detection branch covered. The `CreateDirectory`-UAE case is deliberately skipped — there's no reliable cross-platform trigger for UAE on directory creation (read-only parents don't block it on Windows, system paths differ under elevated CI); the wrapping is covered via IOException, and the filter itself is now covered directly by the new `Buildvana.Core.Diagnostics.Tests` project (all four classification properties, nesting and `AggregateException` recursion included), which also addresses the Codecov patch hole.
8. All three nits taken: `-reading`/`-writing`/`-appending`/`-opening` renames, Abstractions `Description` refreshed, `TempFile.Dispose` guarded.

Sanity gate re-run: build and `bv pack` at zero warnings, ReSharper inspectcode at zero results, all tests green.

rdeago:

rdeago:
All items addressed, one commit per finding (01e95e3…7a7ec0b):

1. **Fixed** (01e95e3) — `FindSolutionFile` goes through `UserDirectory.EnumerateFiles`; the case-insensitivity delta is deliberate. `serializer.OpenAsync` is now in the PR body's exclusion list as a third-party boundary.
2. **Fixed** (1fe8239), with a different trigger than suggested: the directory-as-target trick can't reach this branch — both loaders gate the read on `File.Exists`, which reports **false** for a directory, so a directory named `buildvana.json` short-circuits to "empty config" / "does not exist" without touching the guarded read. The tests hold the file open with `FileShare.None` instead: `File.Exists` (an attribute check) stays true while `File.ReadAllText` fails — portably, since .NET maps `FileShare` to advisory locks on Unix.
3. **Deferred to #330** — the facts are right, but `PrepareForRelease` and `UpdateNewSectionTitle` are untested state machines, so the issue pairs the `ReadAllLines` switch with a ChangelogService test suite, tests first.
4. **Accepted deliberately**, now recorded in the PR body. Confirmed `Microsoft.Extensions.FileSystemGlobbing.dll` (47 KB) is in the nupkg; next to the multi-megabyte LibGit2Sharp natives already in `bin\` it's noise, and a `Buildvana.Core.IO.Globbing` split would be exactly the pre-created project the Core tier layout avoids.
5. Minors: stale `Core.IO` reference dropped (b97b018); `DeleteIfExists` treats a vanished directory as success and uses the sibling `Exists` (1ba9b6c); missing-base-directory behavior pinned with a test and a remarks paragraph (7a7ec0b); PR body refreshed, including the stale `WriteThisAssemblyConstantsFile` bullet; `ToArray()` left alone per your own note.

Sanity gate green: build and inspectcode at zero, all suites pass, `bv pack` produces all three packages.

@claude please review the new commits (01e95e3…7a7ec0b) against your last review.

rdeago:
Both remaining items addressed:

- **Item 5**: added `DeleteIfExists_WithNullPath_ThrowsRaw`, pinning the guard contract that 1ba9b6c had left riding implicitly on `Exists`, plus a `CreateDirectory` twin that completes the raw-throw pins across the `UserDirectory` surface (1c42c85).
- **Item 3 loose end**: the PR description's exclusion list now names `ChangelogService`'s residual mid-read window and points at #330.

Full pre-push gate on 1c42c85: build and tests green (`Buildvana.Core.IO.Tests` now at 82), `bv pack` produced all three packages, `inspectcode` at zero results.

rdeago:
Ci checks skipped because of the current GitHub Actions outage. Our pre-push sanity check is actually stricter than the CI build/test/pack workflow, so this PR can be merged nonetheless.

PR #332, opened 2026-08-07
Delegate non-local bv runs to the repository's pinned version; add bv update

### LINKED ISSUE #325: bv should detect a non-local launch and delegate to the repository's pinned version
### Reference version

Commit c45d5081a8bbefd425b46a16694155356ddeb309 (latest release: 2.1.41-preview)

### Background and motivation

bv, Buildvana.Sdk, and Buildvana.Runtime are a lockstep-versioned group; a repository pins its bv in `.config/dotnet-tools.json` and its SDK in `global.json`. Today the SDK version check _refuses_ to run `usesSdk` commands on a mismatch — a guard, not a remedy: a user running a global or `dnx`-launched bv gets an error and instructions, when what they want is for the right version to just run.

The harm isn't limited to `usesSdk` commands either: `buildvana.jsonc` is written to the schema generated from the pinned version's typed model (`schemas/buildvana.schema.json`), so a bv of any other version deserializes it with a different `Buildvana.Runtime` model — silent drift or spurious validation errors on commands the SDK check doesn't even gate.

The Angular CLI solved this structurally: the global `ng` is a launcher that always delegates to a project-local CLI when one exists, printing "The local Angular CLI version is used." — the user types `ng` anywhere and the repo wins.

### Proposed enhancement

1. bv detects how it was launched — repo-local (tool manifest), global, or `dnx` — or at minimum local vs. non-local.
2. When a non-local bv runs in a repository whose tool manifest pins bv, it delegates the entire invocation to the pinned version: ensures it is available (`dotnet tool restore` if not yet restored), then runs it with the original arguments, inherited stdio, and forwarded exit code, printing an ng-style info line naming the version that ran when versions differ.
3. When delegating to a local version, neither command-line arguments nor the configuration file shall be checked, as they could be compatible with the local `bv` version but not necessarily with the delegating one.
4. A new subcommand fully updates the repository's Buildvana surface in one operation: tool manifest pin, `global.json` SDK pin, config schema reference / `buildvana.jsonc` migration — everything (name open, e.g. `bv update`).

### Implementation proposals

`AppContext.BaseDirectory` distinguishes install modes (global tools execute from `~/.dotnet/tools/.store/…`, manifest-run tools from the NuGet global packages folder, `dnx` from its own cache); these layouts are not contractual SDK surface, so detection must degrade gracefully to local-vs-non-local.

Delegation is `dotnet tool restore` (when needed) + `dotnet tool run bv -- <args>` in the home directory. A recursion guard (e.g. a `BV_DELEGATED` env marker on the child) prevents loops on mis-detection; an opt-out flag mirroring `--skip-sdk-check` covers the rare "run this exact binary" need.

Once delegation lands, the post-update re-run advice and much of `WarnIfToolManifestDisagrees` simplify or become moot. Evaluate whether the `sync-sdk` command is still useful: if not, remove it in the same PR. It's been in `bv` for a very short time and no one is going to miss it if there is a viable or even better alternative, as the full-update command probably is.

### Usage examples

```text
$ bv build                # global bv 2.2.0; repo pins 2.1.58-preview
Delegating to bv 2.1.58-preview from this repository's tool manifest.
[build proceeds with the pinned version]

$ bv update               # name TBD
bv: 2.1.58-preview -> 2.2.0 (tool manifest)
Buildvana.Sdk: 2.1.58-preview -> 2.2.0 (global.json)
buildvana.jsonc: schema reference updated - check 
```

### Risks

- Observable behavior change: plain `bv` in a pinned repo no longer runs the global binary — that's the point, but it needs docs and changelog.
- One extra process hop per delegated invocation.
- `dotnet tool restore` may touch the network and must fail as a clean `BuildFailedException`.
- Detection heuristics can misfire (mitigated by the fallback to local-vs-non-local plus the recursion guard).
- No reliance on CI/runner conveniences, per project policy.

### Additional information

Grew out of PR #321 review round 4, finding 3.

Acceptance criteria: delegation behavior and the full-update subcommand implemented with tests, documentation updated, and `CHANGELOG.md` entries under "Unreleased changes" for every public-facing change.

rdeago:
## Proposed changes

**Why.** The SDK version check was a guard, not a remedy: a global or dnx-launched `bv` in a pinned repository got an error and instructions, when what the user wants is for the right version to just run. And the harm was not limited to `usesSdk` commands: a `bv` of any other version deserializes `buildvana.jsonc` with a different `Buildvana.Runtime` model — silent drift or spurious validation errors on commands the check never gated.

**What.** Like the Angular CLI's global `ng`, `bv` now always defers to the repository:

- **Delegation.** When the tool manifest pins `bv`, a non-local `bv` delegates the entire invocation to the pinned version — `dotnet tool restore` when needed, then `dotnet tool run bv` with the original arguments verbatim, inherited standard streams, and forwarded exit code, printing an info line on standard error when the versions differ. The delegating side neither parses arguments nor reads configuration; judging them is the pinned version's job. `--skip-delegation` runs the exact binary invoked, and the `BV_DELEGATED` environment variable set on the child makes delegation loops impossible.
- **`bv update`.** One operation re-pins the repository's whole Buildvana surface to the running `bv`'s version: the tool manifest entry (via `dotnet tool update` / `install --create-manifest-if-needed`, which also downloads the version), the `Buildvana.Sdk` pin in `global.json`, and the configuration file's `$schema` reference; the configuration is then loaded with the new model and problems surface as warnings. `update` is delegation-exempt ("bring this repository to me") and refuses downgrades without `--force`.
- **`sync-sdk` is removed**, fully subsumed by the two features above; the SDK version check stays as a backstop for repository-internal pin disagreements and now points at `bv update`.
- **`RuntimeInfo`** (formerly `BuildvanaPaths`) now carries the running `bv`'s version and the delegating `bv`'s version (`null` when not delegated) alongside the well-known paths, shared by all hook contexts.

**How.** The delegation decision is primarily a _version comparison_: a manifest-run `bv` always matches the pin by construction, so a mismatched `bv` cannot be the manifest's, and the delegated child never re-delegates. Install-layout detection (`AppContext.BaseDirectory` matched against the tool-store and package-cache layouts, degrading to unknown ⇒ non-local) decides only the equal-version case, so the manifest's install always runs no matter how `bv` was launched. Delegation needed true stdio inheritance, which CliWrap cannot do, so `IProcessRunner` gains `RunWithInheritedStdioAsync` implemented with raw `System.Diagnostics.Process` (parent survives Ctrl-C; the child owns shutdown). A new `docs/EnvironmentVariables.md` inventories the variables `bv` reads and sets.

Verified end to end on this repository: a dev-built `bv` delegates `--version` to the pinned 2.1.70-preview, and `--skip-delegation` runs it in place.

- [x] Fixes #325
- [ ] Partially fixes #

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Breaking changes

- [x] Yes
- [ ] No

Relative to the 2.1 preview line only (no stable release is affected): plain `bv` in a pinned repository no longer runs the invoked binary; `sync-sdk` is gone; `BuildvanaPaths` is reshaped into `RuntimeInfo` and the hook context's `paths` member becomes `runtimeInfo`.

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Point-by-point follow-up on [the review](https://github.com/Tenacom/Buildvana/pull/332#issuecomment-5210240445); changes are in f2c4ff4.

**1 — Adopted, with one correction to the scenario.** `"2.1"` actually parses (`NuGetVersion` accepts partial versions); a genuinely unparseable version makes the manifest unusable to the dotnet CLI itself — `restore`, `update`, and `install` all fail to read it, since the CLI parses manifest versions with the same `NuGetVersion.TryParse` bv calls (`ToolManifestEditor` in dotnet/sdk). So switching verbs can't rescue that repository; what bv owes the user is a clear error instead of a forwarded CLI parse failure. `ReadBvPin` now reports entry-presence, raw version text, and parsed version separately; `bv update` fails up front naming the entry to fix or remove; the verb choice is keyed on entry presence; and delegation warns instead of silently running in place. Both unusable shapes are tested on both consumers.

**2 — Adopted, in a stronger form than either suggested alternative.** "Run first, restore on failure" can't work under inherited stdio: a resolver miss and the delegated bv's own non-zero exit are indistinguishable to the parent, so every failed build would restore-and-retry. Instead, the restore is now gated by a probe of the SDK's tool resolver cache — which is exactly and only what `dotnet tool run` consults: an entry matching the pinned version whose `PathToExecutable` exists (`LocalToolsCommandResolver` / `LocalToolsResolverCache` in dotnet/sdk; NuGet package presence is neither necessary nor sufficient). The cache is internal SDK surface, so the probe is trusted only in the skip direction — a miss, an unreadable file, or format drift falls back to restoring, i.e. to the prior behavior; the one residual window (the TFM dimension, unknowable without a process spawn) is documented in the class remarks. On a miss, the restore now streams its output to stderr, and a failed restore degrades to a warning followed by the run attempt: `dotnet tool restore` attempts every manifest tool even when one fails, so an unrelated tool's unreachable feed (this repo's `ngbv`, say) no longer blocks a delegated `bv build`, and when bv itself is the missing one, `dotnet tool run` fails visibly with its own actionable message. Net: the common delegated invocation costs one process spawn, and the coupling to unrelated tools' feeds is gone.

**3 — Keeping the subsystem; the counter-scenario is now in the doc comment, per your fallback ask.** Equal versions do not guarantee equal bits: build metadata never participates in version comparison, so a locally built bv carries the same version as the published package it descends from, and a mutable private feed can republish a version outright. Under the repository-authority rule those must yield; only `PackageCache` — the same files `dotnet tool run` would launch, identity by location rather than version-number inference — runs in place. One reframe: the subsystem doesn't enable delegation, it prevents the pointless self-delegation hop on every `dotnet bv` while keeping strict hand-over semantics for everything non-local. With the probe from point 2, the equal-version non-local case no longer pays a restore either.

**4 — Documented** in `docs/DirectoryStructure.md` and the changelog: the home-directory cwd is deliberate (nested-manifest shadowing; `dotnet tool run` has no manifest-path option), bv's own arguments are home-anchored anyway, and forwarded relative paths change interpretation.

**5 — Documented rather than special-cased:** in a delegating repository `--version` answers for the bv that actually runs, consistent with every other judgment the invocation makes; `--skip-delegation --version` asks the invoked binary.

**6 — Declining.** The line prints only on a version mismatch, which makes it a diagnostic — "a different bv than the one you invoked will run" — not decoration; it lives on stderr and cannot dirty piped stdout. `--nologo` is most common in CI, exactly where a silent version switch would be worst.

**7 — Fixed as flagged.**

**8 — Unusable-pin tests added (see 1).** On the cancellation test: `elapsed < 15s` against a 30-second sleep does prove the wait didn't run to completion, and proving actual tree death portably means waiting out the full sleep to assert a "finished" marker never appears — trading a soft bound for a hard 30-second test. Leaving as is.

**Nits — all three taken:** a shared `OwnVersion.Value` accessor replaces the three parse sites; the schema regex is `[GeneratedRegex]` with cross-reference comments on both copies (service and release hook); the downgrade-guard asymmetry is acknowledged in its comment — when a pin is newer the update throws before the schema reference is touched, so only a hand-desynced reference can walk backwards, and it's cosmetic.

rdeago:

rdeago:
All five items are in 277dc09.

**New finding — confirmed and fixed.** Your reasoning held up against the sources you couldn't reach this run: `ToolManifestEditor.GetToolManifestPackageFromOneManifestFile` constructs `new PackageId(toolJson.Name)`, and `PackageId`'s constructor lowercases, so `tool restore` and `tool run` treat a `BV` key as `bv`. `ReadBvPin` now matches the entry case-insensitively, and both consumers have `BV`-keyed manifest tests. One footnote for the record: the CLI is not uniformly case-insensitive itself — `ToolManifestEditor.Remove` compares the raw JSON key ordinally against the lowercased id — but the verbs delegation depends on go through `PackageId`.

**`BV_DELEGATED` inheritance — agreed and closed.** Both of bv's child-spawn funnels (`RunDotNetAsync`, `RunFileBasedAppAsync`) now pass their environment through a helper that layers it over the marker's removal, so builds and hooks no longer carry it; a bv reached through them makes its own delegation decision, and `RuntimeInfo.DelegatingVersion` remains the way a hook learns about delegation. These funnels are the only spawn sites reachable with the marker present — `DelegationService`'s own spawns happen only when the marker is absent, and a delegated bv never runs `update`. `EnvironmentVariables.md` now documents the removal instead of framing the inheritance as a feature.

**The `switch` — reverted.** Ric's verdict: the two `if`s read better; with the early returns the method is linear enough that no simplification was owed. (The switch was ReSharper's suggestion, for what that's worth.)

**The fallback comment — corrected** to say what the code does: an all-empty home degrades to a relative path resolved against the current directory, and a wrong cache directory only ever costs a restore.

**The update ordering — pinned** with a comment stating the rationale: manifest first because it is the one step with an external actor, and the partial-update window self-heals on rerun exactly as you described.

rdeago:

rdeago:
All four round-3 items are addressed, one commit each.

**`bv update --force` vs. the CLI's downgrade guard — confirmed and fixed** (6c6a496). We ran your two-minute experiment against the SDK pinned in `global.json` (10.0.302): `dotnet tool update nbgv --version 3.6.146` over an installed 3.7.115 fails with exit code 1 — _"The requested version 3.6.146 is lower than existing version 3.7.115 (…). Use the --allow-downgrade option to allow this update."_ So the guard does cover the local-manifest path, and forced downgrades died exactly as you suspected. The fix is the conditional argument you outlined: `PinToolManifestAsync` appends `--allow-downgrade` exactly when the manifest change actually is a downgrade — a state only reachable forced, since `EnsureNoUnforcedDowngrade` throws otherwise — so bv passes the flag precisely when it has itself authorized the downgrade, and the CLI's guard stays armed on every other path. The forced-downgrade test now expects the flag, and a new test (`UpdateRepository_WithForce_AndOlderManifestPin_UpdatesWithoutAllowDowngrade`) pins its absence on a forced upgrade, so the flag keys on direction, not on `--force`.

**Delegating bv judges one sliver of the command line** (dad6fd0). Right — the split runs first, and `CliOptionReader.ReadValue` rejects a value-bearing global with nothing after it. We qualified the sentence in `docs/DirectoryStructure.md` rather than catch-and-delegate: delegation consumes the split's outputs (the subcommand for the `update` exemption, `--skip-delegation`), so delegating on a parse failure would need a second, cruder pre-parser to reroute an error message that is identical in every bv version anyway.

**Null-removal contract on the hot path** (192f6b2). Agreed and done: a twin of `RunWithInheritedStdio_RemovesNullEnvironmentEntries` now runs against `RunAsync`, `[NotInParallel]` like its sibling, locking CliWrap's "null removes the variable" semantics in at upgrade time.

**`ReadBvPin` reads one manifest, the CLI reads a chain** (0a3e772). Behavior stays as is — the repository's own manifest is the pin bv manages — but the doc comment now states the deliberately narrower scope explicitly, so the case-insensitivity paragraph's CLI-parity argument can no longer be over-read.

Gates on the pushed tree: the full `bv pack` gate (build, tests, pack) exits 0, and ReSharper inspectcode reports zero results at WARNING severity.

@claude Could you take another look, re-reviewing the range 277dc09..0a3e772?

rdeago:
Both follow-ons addressed:

- **Changelog qualification** (468029d): the delegation paragraph in `CHANGELOG.md` now carries the same qualification as `docs/DirectoryStructure.md` — the delegating `bv` runs the minimal split, and only a value-bearing global option with no following value is rejected before delegation, identically in every version.
- **`ReadBvPin` scope claim** (c8dd79a): the paragraph now says delegation and `bv update` key their _decisions_ on the home directory's manifest alone, while the spawned dotnet CLI follows its own manifest-location rules — landing on the same manifest whenever it has a bv entry, though `bv update`'s no-entry install can reach an ancestor's. Your recollection of the walk-up behavior matched what we verified from dotnet/sdk sources in the previous round.

Gates on the pushed tree: `bv pack` exit 0, ReSharper inspectcode zero results at WARNING severity, all tests passing.

PR #333, opened 2026-08-07
Promote high-value analyzer hints to warnings

rdeago:
## Proposed changes

Following a review of ReSharper's hint-level output, this PR turns the consistently valuable suggestions into enforced rules, codifies the underlying style preference (pattern matching) in the C# style guide, and cleans up everything the new severities flagged. The solution now analyzes clean at _all_ severity levels, hints included.

- **Style guide**: new "Pattern matching" section — prefer patterns over equivalent comparison/type-check/cast chains, with the `Equals`-vs-`==` caveat spelled out.
- **Promotions to warning** (ReSharper keys in `.editorconfig`, plus Roslyn twins IDE0011, IDE0039, IDE0063, IDE0200 where they exist): merge into pattern, use pattern matching, merge cast with type check, `await using`, u8 string literals, `GetValueOrDefault` over `TryGetValue` chains, method groups over closures, local functions over lambdas, init-only properties, braces, simple `using` statements.
- **Test-scoped relaxations** (`tests/.editorconfig`): local-function placement/separation rules, member-can-be-made-static (TUnit test methods must be public instance methods), target-typed `new`.
- **Single source of truth**: test analyzer exceptions moved from `Common.targets` `NoWarn` items to `tests/.editorconfig`, where compiler, IDE, and `inspectcode` all read them.
- **MSBuild task exemption** (`src/Buildvana.Sdk.Tasks/.editorconfig`): get-only/init-only property suggestions are meaningless for engine-set task properties.
- **Resulting code fixes**: test lambdas converted to local functions (via `dotnet format`), fake callback hooks made init-only, one list pattern, plus assorted `await using`/u8/method-group fixes.

No changelog entry: nothing here is public-facing — packages' behavior and public API are unchanged.

Verified: `dotnet bv pack` green (all three packages produced), `inspectcode --severity=WARNING` reports zero results, and a full no-threshold ReSharper pass reports zero diagnostics down to hint level.

- [ ] Fixes #
- [ ] Partially fixes #

_No corresponding issue — analyzer configuration change grown out of an interactive review session._

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] Other

_Other: analyzer severity configuration (`.editorconfig` files, `tests/Common.targets`) and C# style guide update._

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**
  - [ ] I have added tests that prove my feature works / my fix is effective
  - [ ] I have added / modified XML documentation according to changes in code

### DIFF
--- a/.claude/rules/csharp-style-guide.md
+++ b/.claude/rules/csharp-style-guide.md
@@ -48,7 +48,7 @@ Prefer pattern matching over equivalent chains of comparisons, type checks, and 
 Semantic differences to keep in mind when converting:
 
 - List, property, and recursive patterns imply a null test: `propertyPath is [PropertyName, PackageId]` is `false` when `propertyPath` is null, where `propertyPath.Count == 2` would throw. Usually an improvement, but it is a behavior change, not a pure rewrite.
-- Constant patterns compare with `Equals`, while `==` uses the type's equality operator. With `float` / `double` the pattern is what you want: `x is double.NaN` matches NaN, while `x == double.NaN` is always `false`. Exception: in the rare case of a type that overloads `==` with semantics that differ from `Equals` (so `x == null` and `x is null` disagree), keep the operator form.
+- Constant patterns never use the type's `==` operator. Non-null constants compare with `Equals`; `null` is a plain reference test (a `HasValue` check for nullable value types). With `float` / `double` the pattern is what you want: `x is double.NaN` matches NaN, while `x == double.NaN` is always `false`. Exception: in the rare case of a type that overloads `==` so that `x == null` and `x is null` disagree, keep the operator form.
 
 ## Partial classes
 
--- a/.claude/rules/csharp-style-guide.md
+++ b/.claude/rules/csharp-style-guide.md
@@ -24,6 +24,28 @@ Use latest C# features whenever possible. Non-exhaustive list of features to use
 - `field` keyword
 - Primary constructors
 - Extension blocks
+- Pattern matching (see below)
+
+## Pattern matching
+
+Prefer pattern matching over equivalent chains of comparisons, type checks, and casts: a pattern states the shape of the data, while a chain of checks makes the reader reassemble it. In particular:
+
+- List patterns over count checks + indexed comparisons:
+
+  ```csharp
+  // This is wrong
+  var isPinPath = propertyPath.Count == 2 && propertyPath[0] == PropertyName && propertyPath[1] == PackageId;
+
+  // This is correct
+  var isPinPath = propertyPath is [PropertyName, PackageId];
+  ```
+
+- `is T x` over `as` + null check, and over `is` check + cast.
+- Property patterns over chains of dotted comparisons: `result is { ExitCode: 0, StandardError.Length: 0 }`.
+- `is null` / `is not null` over `== null` / `!= null`.
+- Switch expressions over `if`/`else if` chains or switch statements whose branches produce a value.
+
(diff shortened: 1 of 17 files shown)


rdeago:

rdeago:
@claude Applied in 1f30ee0:

- **Point 1** — all seven Roslyn twins promoted to `warning` (IDE0019, IDE0020, IDE0041, IDE0066, IDE0078, IDE0083, IDE0150), with the same style-guide pointer comment the ReSharper block carries. Zero new violations surfaced: the ReSharper-side promotions had already swept the codebase clean. `csharp_style_prefer_extended_property_pattern` stays at `suggestion` — it governs nesting style _within_ a pattern, not whether patterns are used.
- **Point 3** — right on all counts: the five casts were IDE0039-fixer leftovers and compile bare. Removed.
- **Point 4** — reverted; `Act` sits below the arrange-act-assert body again, like its sibling.
- **Nit** — `Act` at `UserFileTests.cs:145` is now `static`.
- **Point 7, in part** — the style guide now names `double.NaN` and notes the implicit null test of list/property patterns.

Not applied, and why:

- **Point 2** — the `convert_to_local_function` promotion stays. With the casts gone, `static void Act() => ...` + `Assert.That(Act)` reads at least as well as the lambda form; the churn you flagged was fixer fallout, not a cost inherent to the rule.
- **Point 5** — `Buildvana.Runtime` has no released stable version; we're in a major preview cycle where binary breaks are free (they don't even make the changelog). From here on the rule only shapes new code, which is what it's for. Worth revisiting after next stable release.
- **Point 6** — keeping the inline suppression. Unlike the globally-disabled family above it, `UnusedMemberInSuper.Global` retains real signal on internal base types, and the inline comment carries the domain rationale an `.editorconfig` line can't. The zero-at-hint-level bar is an IDE habit, not a gate contract — no plans to gate it.
- **Point 7's `-0.0` claim** — checked and false: `(-0.0) is 0.0` evaluates to `true`, and a switch with both `0.0` and `-0.0` arms doesn't even compile (CS8510: second arm subsumed). For zeros, constant `double` patterns behave exactly like `==`; only NaN differs. The operator-overload phrasing also doesn't name a wrong culprit: `null` _is_ a constant pattern, and `x is null` vs an overloaded `== null` (the Unity fake-null case) is exactly an operator-overload divergence. The guide keeps that exception, with NaN as the named primary case.

Post-change gate: build 0 warnings / 0 errors, all tests pass, `bv pack` green, inspectcode at WARNING: 0 results.

@claude please review the latest commit.

rdeago:
Follow-up in 2862bb6, addressing the two actionable leftovers:

- **Point 1 (constant-pattern mechanism)** — applied, with your suggested framing. The bullet now reads: constant patterns never use the type's `==` operator; non-null constants compare with `Equals`, while `null` is a plain reference test (a `HasValue` check for nullable value types). The exception clause now states the observable disagreement (`x == null` vs `x is null`) directly instead of leaning on the old `Equals` premise.
- **Point 5 (duplicate `.editorconfig` key)** — applied. Dropped the out-of-order line 113; the alphabetically placed entry stays. The duplicate predates this PR, but it was a one-line fix on a file already touched here.
- **Points 2–4** — noted, leaving as-is: the guide's preamble already scopes it to rules that can't be analyzer-codified (2); IDE0170 governs nesting style _within_ a pattern and is deliberately left free (3); the tests already use the local-function-from-inception shape that avoids the IDE0039 cast round-trip (4).

Gate: `bv pack` clean, inspectcode at `WARNING` → 0 results.

Issue #335, opened 2026-08-08
Give ReleaseCommand an end-to-end test harness


rdeago:
### Reference version

2.1.70-preview (latest published version); analysis done at commit bcb836cf8f98779020d11d49dd6470376b1c41b7, where the merged line coverage of `ReleaseCommand` is 0/154.

### Background and motivation

Issue #320 extracted the extractable logic from `ReleaseCommand` (hook-args assembly into `PostReleaseHookArgsFactory`, working-tree change tracking into `GitService.TrackChangesAsync`), but what remains is exactly what cannot move out: sequencing and policy. The release commit must exist before the artifact pass; the tag check must precede pack; the post-release commit assembles hook changes and dogfood rewrites; push precedes NuGet push. A regression in that ordering surfaces only during an actual release — still the worst possible moment, since releases are not rehearsable locally. And every PR that touches `ReleaseCommand` accrues a few uncovered patch lines that Codecov rightly flags, with no way to do better until the orchestrator is executable under test.

### Proposed enhancement

`ReleaseCommand.ExecuteAsync` runs end-to-end in a test harness, with only the process and platform boundaries faked, so the sequencing itself is covered by tests.

### Implementation proposals

- Drive the real services over a `TempGitRepo` with a `VERSION` file and a `buildvana.json`: real `GitService`, `VersionService`, `ChangelogService`, `HookRunner`. The composition stays honest — no mock-and-assert-the-mock.
- Fake only the boundaries that already have seams: `IProcessRunner` (`FakeProcessRunner` exists in `Buildvana.Core.Testing`) for `dotnet` invocations, `IFileBasedAppRunner` for hooks, and a test `ServerAdapter` (the class is already abstract with a factory) whose release object records `UpdateRepository` / `AddPostReleaseCommit` / `AddAsset` / `PushUpdates` / `PublishAsync` calls in order.
- Assert observable sequences: happy-path prerelease release; changelog policy variants (`none` | `stable` | `all`, empty-section substitute); hook-modified files joining the post-release commit; version-spec change application; tag-collision failure.
- The test `ServerAdapter` must not deepen the current adapter shape: the platform/git-host split is planned, so the fake should be designed against the seams that split will keep.

### Usage examples

Not user-facing; example of the kind of test this enables:

```csharp
[Test]
public async Task Release_AssemblesPostReleaseCommit_FromHookChangesAndDogfoodRewrites() { /* ... */ }
```

### Risks

[...]

PR #336, opened 2026-08-08
Make the hook-args assembly testable, generalize it for future hooks, and run `bv` from the home directory

### LINKED ISSUE #320: Make ReleaseCommand's post-release hook-context assembly unit-testable
### Reference version

2.1.19-preview (gap highlighted by the Codecov report on PR #319)

### Background and motivation

Since #319, `PostReleaseHookContext` is the typed contract between `bv` and repo-owned post-release hooks. `ReleaseCommand` assembles it inline ([ReleaseCommand.cs#L243-L266](https://github.com/rdeago/Buildvana/blob/83448fe/src/Buildvana.Tool/Subcommands/ReleaseCommand.cs#L243-L266)): well-known paths (with `Path.GetFullPath` anchoring for the artifacts and scratch directories), release/version fields, produced packages, and the dogfooding flag — plus the working-tree snapshot diff that detects which files the hook modified.

None of this is unit-tested: `ReleaseCommand` is a large orchestrator that resolves a dozen services and only runs on cloud builds, so it has no test harness at all. Codecov flags the 14 new context-assembly lines at 0% patch coverage. A mistake in path anchoring, a mixed-up version field (`Version` vs `SemVer`), or a regression in the dirty-files diffing would only surface during an actual release — the worst possible moment, since releases are not rehearsable locally.

### Proposed enhancement

The hook-context assembly and the hook-change detection logic are covered by unit tests. `bv`'s observable behavior does not change.

### Implementation proposals

- Extract the context assembly from `ReleaseCommand` into a small internal factory (e.g., `PostReleaseHookContextFactory` in `Buildvana.Tool/Services/Hooks/`) taking the home directory provider, artifacts path, version information, produced packages, and dogfood flag. `ReleaseCommand` delegates to it.
- Move the "snapshot dirty files around the hook run" logic next to `HookRunner` (e.g., a helper that runs a hook and returns the set of files it modified), so it can be tested with the fakes already used by `HookRunnerTests`.
- Add tests in `Buildvana.Tool.Tests` alongside `HookRunnerTests` and `PostReleaseHookContextTests`: path anchoring (relative artifacts path → absolute), version field mapping including `PreviousVersion = null` for a first release, and modified-file detection (no changes / changes / hook absent).

### Usage examples

Not user-facing; example of the kind of test this enables:

```csharp
[Test]
public async Task CreateContext_AnchorsArtifactsPathToWorkingDirectory() { /* ... */ }
```

### Risks

Refactor-only; no public surface or behavior change, so no changelog entry needed. Main risk is cosmetic churn in `ReleaseCommand`; the extraction is mechanical.

### Additional information

Flagged as a non-blocking note in the [PR #319 review](https://github.com/Tenacom/Buildvana/pull/319#issuecomment-5170819562); acknowledged as a pre-existing gap in [this reply](https://github.com/Tenacom/Buildvana/pull/319#issuecomment-5171403293).

rdeago:
## Proposed changes

Three stacked changes, one commit each:

1. **Make the post-release hook-args assembly unit-testable** (#320): the args assembly moves from
   `ReleaseCommand` into `PostReleaseHookArgsFactory`, and the working-tree snapshot diff into
   `GitService.TrackChangesAsync`, a general "run an operation and report the files it modified"
   primitive. Both are fully covered by tests; the 14 lines Codecov flagged on #319 now live in
   tested types, and `ReleaseCommand` retains only orchestration (its harness is filed as #335).
2. **Generalize the hook-args model ahead of future hooks**: `RuntimeInfo` and the args-file loading
   plumbing hoist into a `HookArgs` base record; the new `IHookEvent` interface carries each hook's
   context/event as static abstract properties, so `HookRunner.RunHookAsync<TArgs>` dispatches a
   hook from its args type alone and a mismatched identity is unrepresentable;
   `PostReleaseHookArgsFactory` rebases onto `HookArgsFactory<TArgs>`, which owns the `RuntimeInfo`
   assembly shared by every hook. Done now, while `Buildvana.Runtime`'s surface is young, because
   every hook built on the flat shape would make this refactor more expensive.
   `[JsonPropertyOrder(-1)]` keeps `runtimeInfo` first in args files for human readers.
3. **Run `bv` from the home directory**: an `AnchoringHomeDirectoryProvider` decorator makes the
   home directory the process's current directory at the moment discovery succeeds. This closes a
   latent class of bugs (e.g. `bv release` from a subdirectory resolving `artifacts/<Configuration>`
   against the wrong root) and unifies non-delegated runs with the documented delegated-run
   semantics. Hooks running from the home directory is now a contract-level guarantee in
   `docs/Hooks.md`. Discovery failure leaves the current directory untouched, so commands that run
   outside a repository are unaffected.

Verification: `dotnet bv pack` green (0 warnings, 0 errors), 287/287 Tool + 12/12 Runtime tests,
ReSharper solution-wide inspection at WARNING severity: 0 results. New code is fully covered except
the ~12 orchestration lines in `ReleaseCommand` (see #335).

Behavior note: on the preview line, non-delegated runs invoked from a subdirectory now resolve
relative forwarded arguments against the home directory (previously the invocation directory);
documented in the changelog's subdirectory-invocation entry.

- [x] Fixes #320
- [ ] Partially fixes #

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Resolved in four commits on top of the rebase: `c019f13`, `b8ba7dc`, `af50dea`, `374f405`.

**1. Documented ordering — confirmed, and it had a third casualty.** `git log -S` pins the cause: `ac9567a` reordered the code and left the prose behind. `docs/Hooks.md` and the changelog now say the hook runs _before_ the built-in rewrites, and why: the hook is a file-based app inside the repository tree, so building it resolves the repository's own version pins. The review missed one more sentence with the same inversion — the dependency guidance claimed that at hook time `Directory.Packages.props` "has already been rewritten … restore fails mid-release, deterministically". It hasn't: the pins still carry the previous version, so a versionless self-produced `#:package` restores fine and silently builds against the last release. Subtler trap, now documented as one.

Related: the `Dogfooded` member is now `Dogfooding`. A hook reads it before the rewrites run, so the name asserted something false; the value is unchanged — the resolved outcome, `--dogfood` override included, which no hook can obtain from the configuration file.

**2. `artifactsPath` resolution — taken, and generalized.** A new `IHomeDirectoryProvider.GetFullPath` extension in `Buildvana.Core.Abstractions` replaces hand-rolled resolution at all seven sites (three in `HookRunner`, plus `SelfReferenceUpdater`, `SelfVersionService`, `VersionFile`, and the args factory). `CreateRuntimeInfo` no longer reads ambient process state at all, and the test now chdirs elsewhere and asserts the artifacts path still lands under home — the property that actually matters.

Also not in the review: `RuntimeInfo.HomeDirectory` was the one path in the args carrying a trailing separator (discovery normalizes it that way), so a hook comparing it with its own working directory would mismatch. Trimmed now, with a test.

**3. `Directory.SetCurrentDirectory` — taken.** `Buildvana.Core.IO` exists precisely to turn environment-driven failures into `BuildFailedException`, so the call goes through a new `UserDirectory.SetCurrentDirectory`; a test covers a home directory that vanishes between discovery and anchoring.

**Design — the lazy anchor stays, deliberately.** Laziness is what lets `bv --version`, delegating runs, and invocations outside a repository neither fail nor move the current directory; anchoring eagerly would buy an earlier start time by making "not in a repository" the first thing every command reports. That reasoning is now in the class `<remarks>`, with the trade-off and the mitigation (resolve repository-relative paths through the provider, never against ambient state), so it doesn't get relitigated from scratch.

**Minor.** `Load()` without a home directory: test added — it covers the branch and pins the "hooks run from the home directory" contract from the Runtime side. Trailing-separator comment: added. `ThrowingHomeDirectoryProvider`: moved to `Buildvana.Core.Testing`, and changed to fail from `Resolve` like real providers do (as a public class, the throwing property tripped CA1065). The extra working-tree scan when no hook exists: left alone — one libgit2 status against a full clean+build+test+pack isn't worth complicating `TrackChangesAsync`'s contract or duplicating hook-existence logic in `ReleaseCommand`.

Gate: `bv pack` clean, `inspectcode --swea --severity=WARNING` zero findings. Each of the four commits builds and passes its tests standalone.

PR #337, opened 2026-08-09
Close ChangelogService's mid-read failure window and cover it with tests

### LINKED ISSUE #330: Close ChangelogService's mid-read failure window and cover it with tests
### Reference version

2.1.41-preview (last published release); described behavior is as of 7a7ec0b, PR #329's HEAD.

### Background and motivation

#329 (closing #324) made denied or failed user-file accesses fail with a single clean error line. `ChangelogService` opens `CHANGELOG.md` through `UserFile.OpenText`, which — as documented — guards the open only: an I/O failure raised while reading through the returned `StreamReader` still escapes as an unhandled-exception stack trace.

Of the three read sites:

- `HasUnreleasedChanges` stops reading as soon as it finds content in the "Unreleased changes" section; streaming is the right shape there.
- `PrepareForRelease` and `UpdateNewSectionTitle` consume the file to EOF and rewrite it whole, so streaming buys nothing: materializing the read closes the mid-read window at no behavioral cost.

This was deferred out of #329 (re-review, item 3) because both rewrite methods are line-by-line state machines with no test coverage at all; refactoring them to close a narrow failure window without a safety net was a worse trade than leaving them be. The refactor and the missing tests are therefore one work item: tests first, then the switch.

### Proposed enhancement

An I/O failure at any point while `PrepareForRelease` or `UpdateNewSectionTitle` reads `CHANGELOG.md` is reported as the usual single clean error line (`Could not read from CHANGELOG.md: <reason>`), with no other behavioral change; and `ChangelogService`'s parsing and rewriting semantics are pinned by tests.

Acceptance criteria:

- [ ] Tests cover `HasUnreleasedChanges` (empty section, populated section, no-sections failure), `PrepareForRelease` (section move, empty-section substitute, subsection filtering, EOF trimming when the new section is last, LF line endings), and `UpdateNewSectionTitle` (title replacement, single-section failure).
- [ ] `PrepareForRelease` and `UpdateNewSectionTitle` read via `UserFile.ReadAllLines(FileName, encoding)`; `HasUnreleasedChanges` deliberately keeps `OpenText`, with a comment stating why.
- [ ] Changelog entry under _Unreleased changes_ (user-visible failure-mode change).

### Implementation proposals

- Write the tests against the current implementation first, then apply the refactor under them.
- The state machines pivot on `line == null` as the EOF transition; the smallest-diff conversion iterates the materialized lines with a trailing EOF sentinel, keeping the machines' shape. Restructuring the loops is also fine if it reads better.
- Testability notes: the changelog path is relative to the current directory, so tests that redirect it must be `[NotInParallel]`; `MakeSectionTitle` composes the section title from `VersionService`, `ServerAdapter.GetReleaseUrl`, and `DateTime.Now` — assert with a pattern, or add a seam only if pattern matching proves too loose.

### Usage examples

Before, when the read fails mid-file (e.g. a cloud-sync provider yanks the file during `bv release`):

```text
Unhandled exception. System.IO.IOException: The process cannot access the file '...\CHANGELOG.md' [...]
   at System.IO.StreamReader.ReadLine() [...]
```

After:

```text
Could not read from CHANGELOG.md: The process cannot access the file [...]
```

### Risks

- The rewrite methods encode subtle Keep-a-Changelog handling (subsection collection, empty-substitute, EOF blank-line trimming); writing the tests before the refactor is the mitigation.
- Materializing the changelog in memory is negligible at real-world sizes.
- No public API change; no new or changed dependencies.

### Additional information

Agreed in PR #329's re-review (item 3): https://github.com/Tenacom/Buildvana/pull/329#issuecomment-5205291719. Related: #324, #321.

rdeago:
## Proposed changes

`ChangelogService` opened `CHANGELOG.md` through `UserFile.OpenText`, which — as documented — guards the
open only: an I/O failure raised while reading through the returned `StreamReader` still escaped as an
unhandled-exception stack trace. All three entry points now read the file through `UserFile.ReadAllLines`,
so a failure at any point of the read is reported as the usual single clean error line, and all three use
one strict UTF-8 encoding instead of two different ones.

The switch was deferred out of #329 because both rewrite methods are line-by-line state machines with no
test coverage, and #330 pairs the refactor with the missing tests. Covering them turned out to need one
design change beyond what the issue anticipated: `ChangelogService` cannot be constructed from a test at
all, since `ServerAdapter` has a `private protected` constructor and `VersionService` is `sealed` around
a live Git repository. Rather than build seams to mock the environment, the parsing and rewriting logic
moved into a new `ChangelogUpdater` that works on lines already read, leaving file access to the service
— the "extract pure logic out of plumbing" route. The state machines are otherwise unchanged: the only
edits are `reader.ReadLine()` becoming an indexed read with an EOF sentinel, and the section title
arriving as a `Func<string>`.

The title is a factory rather than a string so that it is still composed only when a section is actually
written. Passing a computed string would have moved the call ahead of the "contains no sections" check,
and a malformed changelog would then report a failure raised while building a title nobody will read.
Composing the title is itself a pure function of a version, a URL and a date, so it moved to
`ChangelogUpdater` too, where a test can pin its formatting.

Three behavior changes came out of writing the tests and reviewing them:

- The `release.emptyChangelog` substitute replaces the whole body of the moved section, blank lines
  included, so it used to come out flush against both the new section's title and the heading below it.
  It is now trimmed and surrounded with blank lines of its own.
- A `release.emptyChangelog` that is all whitespace substituted nothing for nothing: `bv release`
  announced the substitution and then wrote a section with no body. It now fails the release with the
  message that says the substitute is missing.

  Neither of the two gets a "bugs fixed" entry: `release.emptyChangelog` is introduced in this same
  unreleased version, so no released `bv` ever had the old behavior, and an entry would offer a reader
  a fix for something they are meeting for the first time a few bullets earlier. The rule is stated
  where the setting is introduced instead — the breaking-change entry now reads "when unset or blank,
  an empty section fails the release".
- The release date in section titles was formatted with the current culture, so a machine set to `th-TH`
  or `ar-SA` wrote a date in a non-Gregorian calendar (`2569-04-27` for `2026-04-27`), matching neither
  the release tag nor the other section titles. It is now formatted with the invariant culture, pinned by
  a test that runs under `th-TH`. This bug predates the PR; it was folded in rather than deferred because
  a follow-up would have cost far more than the fix.

Unrelated to the above, one commit sweeps `BuildFailedException.ThrowIfNot(<negated condition>)` call
sites across the repository into `ThrowIf(<condition>)`. It is committed here because it touches
`ChangelogUpdater`, which this PR rewrites, and would otherwise conflict with it.

A last commit wraps the declarations and attributes of the files this PR works in that exceed the
style guide's line limits, kept apart from the changes under review as the rule prescribes.

The changelog entry for #329 was extended rather than joined by a new one: both changes land in the same
unreleased version, so between stable releases they are one story. The date bug, being the only change
here that a released `bv` actually shipped, has an entry of its own.

- [x] Fixes #330
- [ ] Partially fixes #

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Thanks — findings 1 and 2 were both right, and 1 was right about the changelog entry too.

**1. Mid-read window in `HasUnreleasedChanges`.** Taken. `ChangelogUpdater.HasUnreleasedChanges` now takes `IReadOnlyList<string>` like its siblings, and `ChangelogService` reads through `UserFile.ReadAllLines`. You were right that the "only needs a prefix" justification was about parsing, not I/O, and that `ReleaseCommand` reads the whole file moments later regardless — it is in fact the only caller, so the streaming path bought nothing at all. The repeated `lineIndex < lines.Count ? …` is now a `GetLineOrNull` helper shared by the three state machines.

Two things came along with it, as you predicted: the class remarks (finding 4) are now true as written, and all three methods share one `FileEncoding` field. That last one is a real behavior change, so, stating it plainly: a changelog containing invalid UTF-8 now fails the _check_ with a clean `Could not read from CHANGELOG.md: …` instead of being read with replacement characters and failing at the rewrite. That is the intended reading — a changelog that can't be read faithfully can't be rewritten faithfully — and it falls under the existing #329 entry.

**2. Empty substitute.** Taken, plus one step further. The substitute is trimmed once into a local and an empty result is treated as no substitute. Beyond that, `ReleaseSettings.ResolveEmptyChangelog()` now returns `null` for a whitespace-only value, so `emptyChangelog: "  "` fails the release with the existing actionable message rather than quietly producing a body-less section — the tool was otherwise reporting "substituting the configured text" while substituting nothing.

No `minLength` in the schema: the code now handles the case honestly, and the schema isn't load-bearing at runtime, so the rule would only exist in two places with one of them unable to enforce it.

**3. Vacuous LF test.** Taken — the unreleased section is now empty, so the CRLF-bearing substitute is actually exercised. (The `DoesNotContain("\r")` assertion wasn't entirely idle, since it also pins `writer.NewLine`, but the substitute argument was dead weight and the test name over-promised.)

**4.** Resolved by 1.

**Nits.** `is null`/`is not null` throughout `ChangelogUpdater`; encoding asymmetry folded into 1; added the `UpdateNewSectionTitle` laziness twin. And the `DateTime.Now` culture bug is fixed here rather than deferred — `MakeSectionTitle` goes through `string.Create(CultureInfo.InvariantCulture, …)`, and it got its own bullet under _Bugs fixed in this release_, since it's a distinct user-visible defect. It stays untested: the format lives inside `ChangelogService`, the very type this PR concluded can't be instantiated in a test, and putting a seam behind a one-line date format to assert it would be the coverage theater the repo rules rule out.

Sanity check on the new commit: `dotnet bv pack` clean (0 warnings, 0 errors, all test projects run), `inspectcode --severity=WARNING` at 0 results.

rdeago:
Round 2 addressed. The branch was rebased onto `main` and force-pushed, so the commit range has moved.

**a — the 120-column rule.** `PrepareForRelease` is wrapped. The declaration was 136 characters, not 129.

Your question about `GetLineOrNull` turned out to be the useful part of the finding, so it got answered in the rule rather than in the code. Ric's ruling: the 120-character threshold applies to _declarations_ only — the section is titled "Long parameter lists" — so invocation lines like `BuildFailedException.ThrowIf(...)` were never in scope. What was missing was any limit at all for every other line, which is what left both of us guessing. `.claude/rules/csharp-style-guide.md` now sets a general limit of 140 characters for code and comment lines, and narrows the 120-character rule to declarations. That commit is on `main`, below this branch.

The limit applies to new and modified lines always, and to a whole file when you are actually working in it, as a separate commit. A mechanical sweep touching many files with a line or two each does not count — otherwise the `ThrowIf` sweep would have owed 25 unrelated wraps, and a rule that makes cleanups expensive stops producing cleanups. Under that rule this PR owed three lines, all in files it works in: the 162-character class doc in `ChangelogService` and two lines in `ReleaseSettings`.

**b — the schema description.** Fixed, but not where you pointed. `schemas/buildvana.schema.json` is _generated_ from `[Description]` on `ReleaseConfig.EmptyChangelog`, and `tools/generate-config-schema.cs` runs in check mode in CI: editing only the schema would have made the build fail as stale. Both now read "When omitted or blank", and `--check` reports the schema up to date.

**c — the BOM caveat.** You were right, and it is now verified rather than recalled. A probe writing `EF BB BF C3 28` and reading it back through `File.ReadAllLines(path, new UTF8Encoding(false, true))`:

```text
no BOM:    THREW DecoderFallbackException: Unable to translate bytes [C3] at index 0
UTF-8 BOM: NO THROW -> "?(" [FFFD 0028]
```

A detected BOM does replace the strict encoding with one whose fallback substitutes U+FFFD, so a BOM'd changelog with invalid bytes is read _and rewritten_ with replacement characters. No behavior change here — that is what `main` does too — but the comment at `ChangelogService.cs:31` now says so instead of promising a guarantee the code does not make.

**d — pinning the culture fix.** Taken. `ChangelogUpdater.MakeSectionTitle(string version, Uri releaseUrl, DateTime date)` is pure, needs no seam, and sits next to the `"## " + makeSectionTitle()` that consumes it; the service passes `DateTime.Now`. The parameter is `Uri`, not `string`: `ServerAdapter.GetReleaseUrl` returns `Uri`, and the old code relied on interpolation calling `ToString()`. A `[NotInParallel]` test sets `CultureInfo.CurrentCulture` to `th-TH` and asserts `2026-01-01` — verified failing calendars first: `th-TH` gives `ThaiBuddhistCalendar` → `2569-01-01`, `ar-SA` gives `UmAlQuraCalendar` → `1447-07-12`. My "coverage theater" objection in round 2 was aimed at mocking a clock or a culture provider; this needs neither, so it was the wrong objection.

**One thing you did not flag, and should have caught me on.** Making a whitespace-only `release.emptyChangelog` fail the release is a user-visible behavior change, and it had no changelog entry. It has one now.

The `ThrowIfNot` → `ThrowIf` sweep that arrived on this branch is now described in the PR body, with why it is committed here.

Sanity check on the rebased branch: `dotnet bv pack` exit 0, 0 warnings / 0 errors, 749 tests passed; `inspectcode --swea --severity=WARNING` → `Total results: 0`; schema check up to date.

rdeago:
All three taken. Branch rebased again (the style-guide clarification below is a new commit on `main`), so the range has moved once more.

**1 — the changelog entry.** You are right, and the direction you took it is the one I should have taken myself: the inconsistency was mine, between excusing the blank-line trimming in round 1 and adding an entry for the whitespace fix in round 2. `release.emptyChangelog` ships in this same version, so neither change is a bug any reader ever met. The bullet is gone, and the breaking-change entry that introduces the setting now reads "when unset or blank, an empty section fails the release" — the whole rule, stated once, where the setting is introduced. The PR description says so too, so the two excuses no longer disagree.

**2 — `ReleaseSettings.Parse`.** Ambiguity conceded, and resolved in the rule rather than in this file alone: whole-file cleanup now explicitly carries the declaration limit. "A file brought within the general limit but left with over-long declarations is a file that still owes a second sweep" — which is the whole argument for not splitting the two thresholds.

That widened this PR's debt by one file, since a sweep that ignores declarations is not the sweep the rule now describes: `ReleaseConfig.cs`, which this PR edits one `[Description]` in, had two attribute lines at 158 and 166 characters. Both are split; the arguments stay constant expressions, so `--check` still reports the schema up to date. `Parse` and those two are the last over-limit lines in the files this PR works in, and they are in their own commit, apart from the changes under review.

**3 — typo.** Fixed.

On the `InvariantGlobalization` trap: worth recording in the test rather than in a review thread, so the comment above `MakeSectionTitle_UsesInvariantCalendar_WhateverTheCurrentCulture` now says the test needs ICU data to bite, and what silently happens to it if that property is ever switched on. Whoever flips it meets the warning at the point of failure.

Sanity check on the rebased branch: `dotnet bv pack` exit 0, 0 warnings / 0 errors, 749 tests passed; `inspectcode --swea --severity=WARNING` → `Total results: 0`; schema check up to date.

Thanks for three rounds of it — the two calls you went past a finding to make (the generated schema, this round's changelog reasoning) were both better than the finding.

PR #338, opened 2026-08-09
Extract TextWriterReporter from ConsoleReporter so the reporter tests run in parallel

### LINKED ISSUE #328: Inject ConsoleReporter's output writers so its tests can run in parallel
### Reference version

2.1.84-preview (raised in the [Claude review of PR #327](https://github.com/Tenacom/Buildvana/pull/327#issuecomment-5197527868))

### Background and motivation

`ConsoleReporter` writes directly to `Console.Out` / `Console.Error`, so `ConsoleReporterTests` can only observe its output by swapping the process-global console writers (`Console.SetOut` / `Console.SetError`). That costs three things: the whole test class (currently ~20 tests) is `[NotInParallel]`, a `TUnit0055` suppression is needed because overwriting the console writers can break TUnit's own logging, and every new reporter test inherits the same ceremony through the `CaptureConsole` helper.

`.claude/rules/testing.md` says the preferred answer to hard-to-test plumbing is "a small, honest design change" that makes the code testable — not environment juggling. This is exactly that case, as the PR #327 review pointed out.

### Proposed enhancement

`ConsoleReporter` tests construct reporters bound to private writers and run in parallel, with no process-global state, no `[NotInParallel]`, and no analyzer suppression. `bv`'s observable behavior does not change.

### Implementation proposals

- Add an internal constructor `ConsoleReporter(Verbosity verbosity, bool? colorOverride, TextWriter output, TextWriter error)`; the existing public constructor delegates with `Console.Out` / `Console.Error`.
- Semantics note: today every write reads `Console.Error` at write time, so a later `SetError` would be honored; with injected writers the binding happens at construction. Immaterial for `bv` (the reporter is constructed once in `Program` and nothing swaps the console afterwards), but the remarks should state it. Storing `Func<TextWriter>` accessors instead would preserve write-time binding, at the price of indirection nobody currently needs. (EDIT: although unneeded, this is a correctness issue in `ConsoleReporter`, so I'd lean on doing it. One day `Buildvana.Core.ConsoleOutput` might be promoted to its own package, so it's important that it don't assume `bv` as the only caller.)
- `Buildvana.Core.ConsoleOutput` gains `<InternalsVisibleTo Include="Buildvana.Core.ConsoleOutput.Tests" />` (same pattern as `Buildvana.Tool`).
- `ConsoleReporterTests` drops `[NotInParallel]`, the `TUnit0055` suppression, and the writer-swapping `CaptureConsole` helper in favor of constructing with `StringWriter`s.
- The `DetectColor` / `IsNoColorSet` coverage exclusions stand either way: they read process-global console/environment state by design.

### Usage examples

Not user-facing; the kind of test this enables:

```csharp
using var stderr = new StringWriter();
var reporter = new ConsoleReporter(Verbosity.Normal, colorOverride: false, TextWriter.Null, stderr);
reporter.Error("boom");
await Assert.That(stderr.ToString()).IsEqualTo($"error: boom{Environment.NewLine}");
```

### Risks

Test-design refactor with no behavior change; `Buildvana.Core.*` is unpackaged, so no public surface moves and no changelog entry is needed. The only subtlety is the construction-time vs write-time binding noted above.

### Additional information

Suggested in the [PR #327 review](https://github.com/Tenacom/Buildvana/pull/327#issuecomment-5197527868) as "optional, not for this PR". Opened as its own issue so the follow-up actually exists instead of being politely mentioned by everyone and done by no one.

rdeago:

- [x] Closes #328
- [ ] Partially closes #

## Proposed changes

`ConsoleReporter` wrote straight to `Console.Out` / `Console.Error`, so its tests could only observe it by swapping the process-global console writers. That cost a `[NotInParallel]` on the whole class, a `TUnit0055` suppression, and a `CaptureConsole` helper every new test had to go through.

Rendering now lives in a new abstract `TextWriterReporter`: level labels, color policy, activity nesting and the write lock, rendering to a pair of `TextWriter`s that a derived class supplies. `ConsoleReporter` derives from it and is reduced to the whole of the coupling to the environment — the color decision, and the naming of the standard streams as the writers to render to.

`bv`&#39;s observable behavior does not change. In particular the writers are still resolved at every write rather than captured at construction, so a later `Console.SetOut` or `Console.SetError` is honored exactly as before; with a virtual property this falls out for free.

The tests exercise the base class through `StringWriterReporter`, a new capture double in `Buildvana.Core.Testing`, and touch no process-global state: no `[NotInParallel]`, no `TUnit0055` suppression, no `CaptureConsole`.

**Why not the shape proposed in the issue.** The issue proposed an internal constructor taking the two writers. A base class was preferred because:

- it makes the environment/logic split structural rather than documented — after the split, _every_ member of `ConsoleReporter` reads or manipulates the process console, so its `[ExcludeFromCodeCoverage]` becomes an honest class-level statement with an invariant attached: anything added there that would deserve a test belongs in the base instead;
- write-time binding needs no machinery. The injected-writers design had to store `Func` accessors to keep it, an indirection nobody would otherwise want;
- whether to emit color is a property of the destination, so the derived class settles it. An injected-writers constructor still taking `bool? colorOverride` would auto-detect against — and, when forced on, mutate the mode of — a console it does not write to.

The `InternalsVisibleTo` grant the issue expected to remove stays regardless: `VirtualTerminalTests` covers the internal `IsNonWindowsTerminalCapable`.

Two names in the diff are not free choices: the writer properties are `OutputWriter` / `ErrorWriter` because CA1716 rejects a virtual member named `Error`, a reserved keyword in Visual Basic; and `ConsoleReporterTests` becomes `TextWriterReporterTests`, after what it now tests. `Buildvana.Core.Testing` cross-references its two reporter doubles in XML docs so the next caller picks the right one: `CaptureReporter` records reporting _calls_, `StringWriterReporter` records rendered _text_.

Activity line formatting is shared rather than duplicated. `TaskLoggingHelperReporter` carried its own copy of `FormatActivityLine`, under a comment asserting it produced the same output as `ConsoleReporter`&#39;s — a claim nothing pinned, and whose referent this PR moved. Both reporters now call `ActivityLineFormatter` in `Buildvana.Core.Abstractions`, which is where the reporting contract already lives and the only reporting assembly `Buildvana.Sdk.Tasks` references. Splitting the old nullable-`elapsed` method into `FormatStart` and `FormatOutcome` drops a discriminator both call sites already knew statically, and the elapsed time gets its first tests: it comes from a `Stopwatch` inside an open scope, so it could never be pinned through a reporter.

No changelog entry: `Buildvana.Core.*` is unpackaged, so no packaged public API moves.

## Additional changes

- The activity line formatter is rewritten from `string.Format` to interpolated strings, with the elapsed seconds still formatted invariantly. The code was only being moved, so this goes beyond the issue; the placeholder pair that made the outcome message&#39;s separator optional was the least readable line in the file, and it now reads as a named local.
- The `NO_COLOR` explanation on `IsNoColorSet` becomes a plain comment instead of an XML doc comment, matching the other private helpers left in `ConsoleReporter`.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

  - [x] I have read the Code of Conduct and agree to be bound by it
  - [x] I have read the Contributor Guide and agree to follow it
- **For code changes only:**


rdeago:

rdeago:
Thanks — findings 1, 3 and the `<Description>` nit are fixed in `beec48b`; the rest are deliberate.

**1 — stale reference, and the divergence behind it.** Fixed at the root rather than in the comment. Both reporters now call a shared `ActivityLineFormatter` in `Buildvana.Core.Abstractions` — the assembly that already holds `IReporter`, `IActivityScope` and `Verbosity`, and the only reporting assembly `Buildvana.Sdk.Tasks` references (pointing it at `Buildvana.Core.ConsoleOutput` would drag `CommunityToolkit.Diagnostics` and the `VirtualTerminal` P/Invokes into the task assembly for four lines of string building). The nullable-`elapsed` discriminator is gone with it: `FormatStart` and `FormatOutcome` are separate methods, since both call sites already knew statically which line they wanted. `ActivityLineFormatterTests` pins the format, elapsed rendering included — which had no coverage at all before, because through a reporter the elapsed time comes from a `Stopwatch` inside an open scope. No follow-up issue: the duplication is gone, so there is nothing left to track.

**2 — `ConsoleReporter`&#39;s stream mapping.** Accepted as-is, deliberately. The only test that can observe the mapping has to swap `Console.Out` / `Console.Error`, reinstating the `[NotInParallel]` and the `TUnit0055` suppression that #328 exists to remove, in order to assert that `Console.Out` is `Console.Out`. `testing.md` names that shape: "do NOT create abstractions whose only purpose is to let a test mock the environment and assert the mock". The `[ExcludeFromCodeCoverage]` isn&#39;t hiding a hole so much as stating the contract — the standard streams are the environment, which is what the attribute is reserved for here.

**3 — one-directional cross-reference.** Fixed: `CaptureReporter` now carries the mirrored `<remarks>`.

**4 — `depth` and invariant formatting.** Left alone, code and comment both, as you suggested. `ActivityLineFormatterTests` does include a case that formats under `it-IT` and still expects `2.5`, so the part that could actually break is pinned rather than explained.

**5 — nits.** `<Description>` updated. The transitive-reference spread stands as designed. On the color remark: `StringWriterReporter`&#39;s destination is a string buffer whose meaning the test defines, so the test is the thing that knows whether escapes belong in it — the base&#39;s claim holds.

Sanity check on the result: `bv pack` clean (0 warnings, 0 errors), ReSharper `inspectcode --severity=WARNING` reports 0 results, `Buildvana.Core.ConsoleOutput.Tests` at 61/61.

@claude please review again.

rdeago:
Both taken, in `cfcad63` and `021edce`.

**a — the culture comment.** You're right, and the test turns out to be safe for a stronger reason than it claimed. Verified against `CultureInfo.cs`: the getter reads a `[ThreadStatic]`, but the setter writes `s_asyncLocalCurrentCulture`, and it's that `AsyncLocal`'s `AsyncLocalSetCurrentCulture` handler which maintains the thread-static — so the value is ambient to the executing flow, and the swap can't reach a sibling test even across an await. The comment now says so, and adds that the `finally` is hygiene for the remainder of this test rather than what keeps the others clean.

**b — the rule text.** Folded in rather than deferred. The `.Abstractions` discipline now reads "usually extension methods on contracts, but free-standing static helpers where there is no contract to hang them off", followed by the criterion that keeps the allowance from becoming a dumping ground — the helper's job is to make every implementation of a contract behave identically in some respect, while its inputs are plain values rather than the contract itself — with `ActivityLineFormatter` as the example. Separate commit from (a): the two points are unrelated.

Sanity check on the result: `bv pack` clean (0 warnings, 0 errors), ReSharper `inspectcode --severity=WARNING` reports 0 results, 758/758 tests passing.

PR #340, opened 2026-08-10
Review code style

rdeago:

- [ ] Closes #
- [ ] Partially closes #

## Proposed changes

Fixed some ReSharper hints.

Not all hints were taken: some did not improve readability, some even made the code noticeably _less_ readable.

**No behavior change is expected.**

## Additional changes

- `ParseBool` was duplicated, byte for byte, in `ReleaseSettings` and `VersionAdvanceSettings`. Both copies are gone, replaced by `CliOptionReader.ReadBoolValue`, which reads the option and parses it in one call. Same diagnostic message, so no behavior change. Covered by new tests in `CliOptionReaderTests`.
- Retuned seven ReSharper hint severities in `.editorconfig`, so that this kind of review is enforced by the gate rather than repeated by eye:
  - promoted to warning: `ConvertToConstant.Local`, `CanReplaceCastWithLambdaReturnType`, `RedundantLambdaSignatureParentheses`;
  - suppressed: `ArrangeObjectCreationWhenTypeNotEvident`, plus `UseRawString`, `UseVerbatimString` and `RawStringCanBeSimplified` as a group — those three disagree with each other by design, and the judgment they were nagging about is now written down as a new **String literals** section in `.claude/rules/csharp-style-guide.md`.
  
  Promoting `ConvertToConstant.Local` immediately caught four locals in `SelfVersionServiceTests` that the manual pass had missed; they are now `const`.
- Made `.editorconfig` the only place ReSharper inspection severities are set. `Buildvana.slnx.DotSettings` held six more, one of which (`UnusedMember.Global` at `WARNING`) contradicted `.editorconfig`'s `none` outright. `.editorconfig` wins that race, so the entry was already dead; the whole `InspectionSeverities` block moved out rather than just that one line, and the file's UI-filter rows were left alone.
- Brought `GitService.cs` fully within the line-length limit. The branch had already reworked that file, and the style guide asks for the leftover wraps as their own commit once that is the case.
- Fixed `.claude/tools/inspect.cs` exiting zero on a _failed_ build in its non-gate modes. Those modes deliberately do not fail for findings, but a failed build is not a finding: nothing was analyzed. Since the VS Code task's problem matcher requires a file position, a build failure whose diagnostics carry none left the Problems panel empty behind a successful-looking task. A failed build now exits 2, exactly as a failed `inspectcode` run already did; a build that merely reported diagnostics still exits zero. The `--gate` mode was already correct and is unchanged.

## Types of changes

- [ ] Bug fix 
- [ ] New feature 
- [x] Changes to repository files (`.gitattributes`, `.gitignore`)
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**
  - [ ] I have checked that all the links I added or modified in XML documentation point to their intended destination


rdeago:

rdeago:
Thanks — good review. Took six of the seven findings; the one I deviated on is #5, where I think there's a better home than the one proposed. All eight commits are pushed, and the sanity-check gate (`bv pack` + ReSharper at WARNING and above) is green: 0 errors, 0 warnings, 0 infos.

### Addressed

**1. `GitHubServerAdapter` — `_token` restored.** Done, and the argument is stronger than the review put it: `ServerAdapter.PushPassword` is declared `string?`, and `GitLabServerAdapter` returns `null` for it. So the base contract genuinely is "push credential, may not exist" — sourcing the REST API credential from it worked only because GitHub happens to use one token for both. `CreateGitHubClient` reads `_token` again, and the declined `ConvertToAutoProperty` hint now carries a suppression comment with the reason, so a future sweep doesn't silently redo it. (`f7d01a1`)

**2. `ExceptionExtensions` — parentheses restored.** Agreed, the edit-fragility argument is the decisive one. No suppression comment needed: `ArrangeRedundantParentheses` is Hint severity, below the gate's threshold, and the XML doc directly above already states the intent. (`69db60e`)

**3. `GitService.cs:218` — wrapped.** One correction to the review: that file has **eight** lines over 140 (41, 88, 120, 125, 209, 218, 263, 269), not just 263. Your mechanical-sweep read is right, though — 18 files at a few lines each is exactly that carve-out — so only 218 is wrapped here, and the rest is left for a standalone cleanup. (`29014c6`)

**4. `StringBuilderExtensions` summary — added**, wording matched to the `Buildvana.Core.Process` twin. (`6ce8ba9`)

**6. Both guard clauses reverted.** Agreed — no de-nesting, three lines each. Both files are now byte-identical to `main`. The `GitService` one stays, for the reason you gave. (`fbeac9a`)

**Nit — `return root`.** Done. (`1ba0221`)

### Deviated

**5. `ParseBool` — extracted, but to `CliOptionReader`, not to a helper beside the settings classes.** Agreed it should be fixed now rather than filed, but a free-standing helper in `Buildvana.Tool.Subcommands` leaves both call sites still naming each option twice:

```csharp
CheckPublicApi = ParseBool(reader.ReadValue("--check-public-api"), "--check-public-api"),
```

`CliOptionReader` already owns option-token semantics and already throws `BuildFailedException` for bad option usage, so the parsing belongs there. It's now `CliOptionReader.ReadBoolValue`, and the call sites read `CheckPublicApi = reader.ReadBoolValue("--check-public-api")`. Both copies are gone, the diagnostic message is unchanged (both call sites passed the long name as the option name), and there are five new tests in `CliOptionReaderTests`. Recorded under _Additional changes_. (`c839bee`)

### One more commit

`c0093c7` fixes my own new tests: `TUnitAssertions0015` rejects `IsEqualTo(true)`/`IsEqualTo(false)` in favour of `IsTrue()`/`IsFalse()`. Caught by the gate on the first run, fixed in its own commit rather than an amend.

### Not verified

You couldn't build; I could. The gate ran the full clean + restore + build + test + pack chain plus ReSharper, all green, so the compile-level claims in your review are now confirmed rather than read.

@claude please re-review.

rdeago:
One more commit since the re-review request, so it isn't pointed at a stale head: `8cfa82d` retunes seven ReSharper hint severities in `.editorconfig`, plus a new **String literals** section in the C# style guide. Gate is green. Details are under _Additional changes_; the reasoning is worth a paragraph here since it changes what future reviews of this kind will and won't have to catch by eye.

**Promoted to warning** — `ConvertToConstant.Local`, `CanReplaceCastWithLambdaReturnType`, `RedundantLambdaSignatureParentheses`. All three are mechanical, with no legitimate exception, so the gate is a better place for them than a manual pass. `ConvertToConstant.Local` earned its promotion immediately: four locals in `SelfVersionServiceTests` had been missed and are now `const`.

**Suppressed** — `ArrangeObjectCreationWhenTypeNotEvident`, and the three literal-form inspections as a group.

That group is the interesting one. `UseRawString`, `UseVerbatimString` and `RawStringCanBeSimplified` disagree with each other by construction: promoting one puts it in tension with the next, and the right answer genuinely varies. Raw wins where the content carries quotes — the `SelfReferenceUpdater` regexes in this PR are the case in point. Escapes win where an escape _is_ the point: `"{\n  \"name\": 42\n}"` in a test asserting line and column keeps its newlines independent of the file's line endings, where a raw literal would make them a property of the source file. So rather than gate any of them, the judgment is now written down in `.claude/rules/csharp-style-guide.md`, and `JsonSchemaValidatorTests.cs:64` keeps its `"""{}"""` for consistency with the literal it is paired with.

Two notes on things I checked rather than assumed:

- A misspelled `resharper_*_highlighting` key is silently ignored, which would make a promotion a permanent no-op. I verified the suppressions by confirming the two inspections that were actually firing disappeared from a HINT-severity run (44 hints → 42), and verified `RedundantLambdaSignatureParentheses` — which has no live sites — by temporarily adding a redundant-paren lambda and confirming it reported at **warning** rather than its default hint. Probe reverted.
- The `RedundantPatternParentheses` hint behind finding #2 also fires on `Buildvana.Runtime/ExceptionExtensions.cs:29`, the BCL-only twin of the file I restored parentheses in. That copy had kept its parentheses all along, so the sweep had de-synced the two and the restore re-aligned them. Both stay at hint.

@claude please re-review — the diff has grown since the last request.

rdeago:
`b0eda36` brings `GitService.cs` fully within the line-length limit — the seven lines left over after the targeted fix in `29014c6`. The style guide asks for exactly this once a branch has actually worked in a file, and for it to land as its own commit, which it does. Gate is green.

Nothing subtle in six of them: two invocations wrapped one argument per line, a lambda ternary split across three lines, three doc-comment paragraphs rewrapped, an object initializer expanded, and the compound path check in `Stage` moved into an `isOutsideRepo` local per the "no multi-line conditions" rule.

The seventh is worth calling out, since it is the one judgment call in the commit. It was a URL:

```
// https://github.com/libgit2/libgit2sharp/blob/5085a0c6173cdb2a3fde205330b327a8eb0a26c4/LibGit2Sharp.Tests/PushFixture.cs#L183-L187
```

144 characters, and unwrappable — splitting a URL across lines breaks click-through, which is the entire value of citing it. Rather than leave it over the limit or damage it, I abbreviated the 40-character commit SHA to 10. GitHub resolves abbreviated SHAs in blob URLs and redirects to the full commit, so the link still works and still pins the same revision, at 114 characters. Ten hex digits is well clear of collision range for that repository. Easy to revert to the full SHA if you would rather have the limit broken than the SHA shortened — the line is the only thing that changes either way.

@claude no new review needed for this one unless you want to look at it — it is mechanical wrapping on top of the diff you are already reviewing.

rdeago:
Both reviews addressed; `07bad9d` is the head and the gate is green.

**First, a correction to my own record.** My `b0eda36` message and the comment above it both say "two invocations wrapped" and "three doc-comment paragraphs rewrapped". It was **one** invocation and **two** doc paragraphs — I counted the `Commands.Stage` wrap, which belonged to `29014c6`. The seven lines are all accounted for either way, so the sweep is complete; only the prose was off. The commit message stays as pushed rather than being rewritten.

### Addressed

- **`<returns>` on `ReadBoolValue`** (`a8e326e`) — it described the raw token, inherited from `ReadValue`. Now describes the parsed value.
- **The unexercised promotion** (no commit) — probed as suggested: restored the cast at `JetBrainsAnnotationsExporter.cs:69`, confirmed it reports at **warning** rather than hint, reverted. `resharper_can_replace_cast_with_lambda_return_type_highlighting` is spelled correctly and live. All three promotions are now verified rather than assumed, which was the point.
- **The self-contradicting bullets** (`f757f07`) — correct, and it was the worst kind of doc bug: the section is the only arbiter now that the three inspections are off, and it disagreed with the repo's own seven verbatim literals. Quotes are now the sole discriminator, with a sentence saying why backslashes are not one, since raw and verbatim treat `\` identically.
- **The two-mechanism conflict** (`07bad9d`) — resolved by removing severities from `Buildvana.slnx.DotSettings` entirely rather than deleting the single contradicting line, which is Ric's call and the one that prevents a rerun of this discussion.

### On which mechanism wins, since the finding left it open

Measured rather than reasoned: flipping `.editorconfig`'s `resharper_unused_member_global_highlighting` to `warning` surfaces **38** `UnusedMember.Global` instances, while the committed `none` leaves the gate green. So `.editorconfig` overrides the DotSettings entry, and that entry was already dead — deleting it is a no-op, while deleting the `.editorconfig` line would have reddened the gate with 38 warnings. Not symmetric, which is what made the choice easy.

The other five moved across keeping their severity: the four `Annotate*` promotions, and `ConvertToPrimaryConstructor` as `none` (`DO_NOT_SHOW`'s equivalent). `UnusedMember.Global` was dropped as a duplicate. UI-filter rows untouched, per Ric — including the three now-stale ones, which cost nothing.

Same spelling risk applied to the migration, so it got the same treatment: a misspelled key would restore `ConvertToPrimaryConstructor` to its default severity, and this codebase has no shortage of primary-constructor candidates. It stays absent from a HINT-severity run, so the key took effect. The four `Annotate*` keys have no site in the repo that triggers them, so their spelling is unverifiable by probe — worth stating plainly rather than implying the whole migration was measured. They follow the same mechanical PascalCase-to-snake_case mapping confirmed four times over by now, and at zero sites a typo would be inert either way.

### One thing I introduced and am leaving

The `isOutsideRepo` extraction in `b0eda36` created a new `ConvertIfStatementToReturnStatement` hint at `GitService.cs:217` (2 sites → 3). Taking it would mean `return isOutsideRepo ? throw … : pathInRepo`, and the style guide rules out ternaries used for side effects. Hint severity, so it cannot reach the gate. Left alone deliberately.

PR #341, opened 2026-08-11
Give ReleaseCommand an end-to-end test harness

### LINKED ISSUE #335: Give ReleaseCommand an end-to-end test harness

rdeago:

- [x] Closes #335

## Proposed changes

`ReleaseCommand.ExecuteAsync` now runs end to end under test, over a real Git repository in a temporary directory, with only the boundaries faked: child processes (`FakeProcessRunner`), the hook's file-based app (`FakeFileBasedAppRunner`), and the server (a recording `ServerAdapter`). The service graph, the versioning, the changelog, the commits and the push are the real ones. Line coverage of `ReleaseCommand` goes from 0/154; 36 tests were added.

**Composition.** `Program.BuildServiceProvider` was private inside an `[ExcludeFromCodeCoverage]` class, so the graph every command runs on was unreachable from a test. Its registrations moved verbatim to `ServiceCollectionExtensions.AddBvServices()`; `Program` keeps the five singletons a host decides for itself. The harness calls the same method and registers its fakes after it, where the last registration of a service type wins — so the composition under test is the real one, not a copy that drifts.

**Sequencing is asserted from effects, not from calls.** Every recorded step carries the repository's state at the moment it happened, so "the release commit existed before the artifacts were packed" and "the branch was pushed before the packages were" are read off the repository. The fake server adapter carries no logic and its knobs are grouped by whether they will belong to the CI platform or to the Git host, so the planned split can cut it in two without rewriting what it answers.

**`Buildvana.Core.Testing`.** `TempGitRepo` gained a pushable bare remote, tagging, committer identity, and history reporting. Two defects surfaced while building it, both fixed here: the default log sort is by timestamp, so commits made within the same second came back in arbitrary order (now topological); and pointing libgit2's configuration search paths per instance mutated a global underneath repositories other threads held, which crashed the test host with a corrupted heap (now a type initializer, once per process). The isolation also makes every `TempGitRepo`-based test independent of the machine's global Git configuration, which previously decided whether a committer identity existed at all — laptop and CI runner took different code paths.

## Additional changes

The harness found three things the issue did not foresee. Each is its own commit.

- **`bv release` refused an impossible release only after building it.** The preliminary checks — cloud build, branch, public-release branch, committer identity — and the version-spec computation ran _after_ the verification pass, so a release that could never succeed cost a full clean, build and test cycle first. They now run before it. (`ReleaseCommand`)
- **A version-spec change published a version one patch below its own artifacts.** `--bump minor` on the 2.3 line tagged `2.4.0-preview` while the commit it tagged built as `2.4.1-preview`: `UpdateRepository` staged the version file into the release commit _after_ the version had been computed from that commit, and the Git height is computed from committed content, so the new version line looked as if it had no committed history — height 0, which is not a height at all. The packages were built, attached and pushed as `2.4.1-preview` under a `2.4.0-preview` tag, and produced-package discovery matched none of them, silently skipping the dogfood rewrites. `NameReleaseCommit` now holds the rule in one place: the commit is made under a provisional message and named afterwards, and every method that changes its tree ends by calling it. (`ServerRelease`, CHANGELOG)
- **Nothing prevented reaching height 0 by another route** — a repository whose `VERSION` was never committed does it without any bump. The final consistency check now rejects it, where the version becomes final and before anything is packed or tagged. Building such a state stays legitimate; only publishing is refused. (`VersionService`, CHANGELOG)

Also: two pre-existing over-long lines in `Program.cs`, wrapped in their own commit per the whole-file line-length rule.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Nine commits, tip `df23a47`. Gate is green: `bv pack` clean, `inspectcode` 0 errors / 0 warnings / 0 info, 357 tests passing.

### Findings

**1 + 2 — three commit objects where one suffices, and `EnsureReleaseCommit`'s stale docs** — `b453071`

Done, as a single commit: they are one change. I had flagged the double amend myself when opening the PR and argued uniformity over a branch; I'm dropping that argument. `Creating release commit...` immediately followed by `Amending release commit...` for one logical operation is a real cost, paid by whoever next reads a release from CI output, and the branch is one `if`. Checked both call sites before taking it: `ReleaseCommand.cs:109` is the creating one, `ReleaseCommand.cs:209` — the post-pack changelog title update — is the amending one, and the early return is correct for both. The docs now state the contract as it stands: the commit is made from whatever is staged, then named.

**3 — mixed cached and fresh handles** — `0c407df`

Done, and one step further than proposed. The rule is a property of the type rather than of two of its members, so it now lives in the type's remarks: members that read state open a handle of their own, and the handle the class keeps is for members that change something. `HeadSha` and `CurrentBranchName` follow it.

**4 — the fake `pack` stamps the version the command believes** — `0ca47cf`

Done; the best catch of the review. `WriteArtifacts` calls `ComputeVersion()` now. The suite passes unchanged, which is exactly the point: what was an assertion in one test is an invariant of every test that packs anything.

**5 — no bump-with-dogfooding case** — `c69ba2f`

Done, asserting on the rewritten files rather than on a recomputed version, as you suggested, so that the post-release commit moving the height stays out of the way. The case fails on the pre-fix code twice over: the release commit carries the wrong version, and the self-references keep the old one because discovery matched nothing.

### Smaller notes

**a — fixed path under the temp directory** — `32144e9`

Done, and I disagree with "close to a wash" in both directions. The argument for changing it is stronger than stated: `Directory.CreateDirectory` adopts an existing directory whoever it belongs to, and the temp directory is shared between users on Linux, so the mechanism that exists to keep the machine's configuration out was one planted `.gitconfig` away from being the thing that let it in. But the leak is not the price either. Nothing may delete that directory while a repository can still be opened — restoring the search paths mid-process is the race the type initializer exists to avoid — which makes process exit the first moment it can go and the last at which any code of ours runs. That is where it goes now; verified that no `bv-test-gitconfig-*` survives a test run.

**b — `Dispose` leaks the remotes when the working tree won't delete** — `9da55e7`. Done, `try`/`finally`.

**c — the remote is recorded before it exists** — `3a61b80`. Done, reordered.

**d — no knob for push credentials** — `98f3469`. Done, and worth having for a reason past coverage: it is the one input the command deliberately treats as non-fatal when missing, and nothing would have objected if that quietly became a hard failure.

**f — `AddBvServices` overstates last-registration-wins** — `df23a47`. Done.

**e — the harness constructor leaks if it throws** — not done. The scenario it guards against is "the harness itself is broken", which surfaces as a failing test on the spot, and the environment variable it would leak is read by nothing outside these same tests. A catch-dispose-rethrow block buys little against that, and every reader of the file pays for it. Say so if you think the failure mode is worse than I read it.

No changelog entry for any of the above: nothing here changes behaviour beyond log wording, and the two release bugs are already recorded.

@claude please review again.

rdeago:
Two commits on top, tip `5daaae0`. Gate is green: `bv pack` clean, `inspectcode` 0/0/0, 359 tests — 357 plus the two argument sets added below.

**The soft edge in the `TempGitRepo` rule** — `88e2f57`

Taken, and by deletion rather than documentation. You are right that "members that change state read what they need through the kept handle" is the missing clause, but writing it down leaves a rule with an exception in it, and the exception is exactly where the stale read lands: a `CreateTag` after a run puts the tag on the wrong commit, which reads as a bug in the code under test. So the handle is gone. Every member opens one and disposes it, the field and its disposal are no longer there, and there is nothing left for the next member to be written against incorrectly. A handle per call costs nothing in a test helper.

**`WithPushCredentials` collapsing two properties** — `5daaae0`

Taken, and the case is stronger than "it gets interesting after the split". It is interesting now: `GitLabServerAdapter` returns `PushUsername => "oauth2"` and `PushPassword => null`, so the mixed shape is what a GitLab release runs with today. The split will not invent that state; it will inherit it.

The option is now the two values themselves, and the warning test runs all three incomplete shapes. What that pins is the `&&` at `ReleaseCommand.cs:73`: were it an `||`, a username alone would count as a credential and libgit2 would be handed a null password. Under the old single knob, that flip changed nothing observable.

**The correction on note c** — recorded, no commit.

Agreed, and thanks for going back to it — the reorder traded one unreachable leak for another rather than closing anything. Making it total means a `catch` around `Remotes.Add` that deletes the directory it just created, i.e. dead code guarding a duplicate remote name that nothing adds. Not worth it; the honest version belongs in the thread rather than in the file.

No new review requested: nothing here touches production code, and you'd already called the PR finished.

PR #342, opened 2026-08-12
Set the console to UTF-8 the way the .NET CLI does

### LINKED ISSUE #339: Investigate bv's and the SDK's behavior on non-UTF-8 consoles
### Context

On Windows, `Console.Out` encodes through the console's active codepage. When that is not UTF-8 (CP850, CP437, CP1252 — the default on most non-English installs), characters outside it are not rejected but best-fit mapped to lookalikes, silently.

Neither `bv` nor the SDK consults the output encoding when deciding what to render. Today nothing breaks, because the reporter emits only ASCII — but that is an accident of the current implementation, not a property anyone is maintaining. The first `✓` or box-drawing character added to a reporter would degrade into something misleading on those consoles, with no error and no test catching it.

### What we know

Measured on CP850, each separately:

| Input | Output |
|---|---|
| `—` U+2014 | `-` U+002D |
| `✓` U+2713 | `V` U+0056 |
| `à ï Ü` | unchanged (present in CP850) |

1. `dotnet bv --version` leaves the console codepage untouched — 850 before, 850 after.
2. A .NET app on a real console at CP850 reports `Console.OutputEncoding.CodePage = 850`.
3. Encoding U+2713 / U+2014 through CP850 is lossy, with best-fit substitution rather than `?`.

To reproduce (3): `chcp 850` in `cmd.exe`, then build a console app whose `Main` is `Console.WriteLine("— ✓");` and run the produced `.exe` **directly** — it prints `- V`. Launching it via `dotnet run` prints it correctly, which is itself a clue worth chasing.

Explicitly **not** in scope: decoding of child process output. That path is sound — CliWrap defaults its `PipeTarget` factories to `Encoding.Default` (UTF-8 on .NET), every child `bv` starts is the `dotnet` muxer, and .NET writes UTF-8 to a redirected stdout regardless of console codepage. Verified under CP850 and documented in `src/Buildvana.Core.Process/ProcessRunner.cs`.

### The question

What do comparable tools do about this, and what should we do? Candidate answers range from changing the console's encoding for the duration of the run, through detecting the encoding and choosing glyphs against it, to emitting only ASCII forever. Each carries different costs, and the right one for a .NET toolchain component is not obvious from first principles.

### Acceptance criteria

- The behavior of every `bv` invocation path (bare apphost, `dotnet bv`, delegated) and of the SDK inside MSBuild is established, not assumed.
- What comparable tools do — the .NET CLI and MSBuild above all — is established from their sources, not from memory.
- A decision is recorded on this issue with its reasoning, so that a future reader can see the trade rather than the conclusion alone.
- Whatever the decision implies is implemented, with a `CHANGELOG.md` entry if it is user-visible.

rdeago:

- [x] Closes #339

## Proposed changes

**Why.** What `bv` could render depended on how it was launched. Reached through the muxer (`dotnet bv`, the form our docs prescribe), the .NET CLI had already switched the console to UTF-8 before `Main` ran and restored it afterwards. Run as a globally-installed shim, nothing had — and characters outside the console's codepage are not rejected but best-fit mapped to lookalikes, so a checkmark arrives as `V`: legible, plausible, wrong. The property that nothing breaks today held only because the reporter happens to emit ASCII, maintained by nobody.

`bv`'s own narration was not the whole of it. Child `dotnet` output is decoded from the pipe at UTF-8 — correctly, whatever the console's codepage — and was then best-fit mangled on the way back out, because `ProcessRunner` deliberately leaves that leg to "whoever owns the console" and, until now, nobody owned it. Every non-ASCII character in an MSBuild diagnostic took that path.

**What.** `bv` now sets the console's output and input encoding to UTF-8 for the duration of its run and restores the previous encodings on exit, honoring `DOTNET_CLI_CONSOLE_USE_DEFAULT_ENCODING=1`. Buildvana SDK is untouched: its tasks run inside MSBuild, which already does this, and under Visual Studio there is no console at all.

**How, and what was rejected.** The issue as filed ruled this approach out on three grounds; the research recorded [in this comment](https://github.com/Tenacom/Buildvana/issues/339#issuecomment-5259633040) found two of them refuted by the .NET CLI's and MSBuild's own sources, and the third — that an encoding gets bytes to a terminal but never gets a glyph into a font — untouched but orthogonal, so it survives as a rule rather than as code. The alternative designs (detect the encoding and choose glyphs against it, or keep output ASCII-only forever) are covered there too, along with why a `bv` that conforms to its toolchain beats a `bv` that decides these questions on the merits.

One deliberate deviation from the CLI: it gates setting both encodings on a non-English UI culture, because that switch exists to render _localized_ CLI output. Buildvana is single-culture English, while the reason for setting the input encoding — without it the change takes effect in `cmd.exe` but not in PowerShell — is language-independent. Copying the gate would leave English consoles behaving differently from every other one.

**Verified by measurement, not inspection.** Same binary, at CP850, with the directory name carrying `✓` so the character reaches a diagnostic as data:

| run | raw output bytes |
|---|---|
| default | `e2 9c 93` — UTF-8 |
| `DOTNET_CLI_CONSOLE_USE_DEFAULT_ENCODING=1` | `cpVtest` — best-fit `✓`→`V` |
| `…=0` | UTF-8 — only the literal `1` opts out |

Codepage is 850 before and after in every case. The control reproduces the original defect exactly, which is the end-to-end confirmation the issue itself could not obtain without redirecting output and destroying the condition under test.

Measured separately, what the console _advertises_ once switched — the reason it is set to a BOM-less UTF-8 rather than to `Encoding.UTF8`:

| encoding passed to the setter | `Console.OutputEncoding.GetPreamble()` | a later `new StreamWriter(stream, Console.OutputEncoding)` writes |
|---|---|---|
| `Encoding.UTF8` | `EF BB BF` | `EF BB BF E2 9C 93` |
| `new UTF8Encoding(false)` | empty | `E2 9C 93` |

The encoded bytes of the character itself are identical in both rows, so console output does not change; only what a later reader of the property inherits does.

**Coverage.** Three decisions are pinned by tests, chosen because each is one a reader could get wrong while leaving the file compiling and the console behaving identically: the opt-out rule (nine cases, including `"1 "` — `set VAR=1 & …` in `cmd.exe` assigns a trailing space); which exceptions mean a console that will not be reconfigured, as against a call that was simply wrong (ten cases); and the absence of a preamble on the encoding handed to the console. Everything else either replaces or restores the process console's encoding, which the test host owns; those carry method-scoped `[ExcludeFromCodeCoverage]` with justifications phrased as why a test cannot honestly reach them. I deliberately did not add a delegate-injection seam to test capture-and-restore: it would exist only to let a test feed fakes and assert on them.

## Additional changes

- `.claude/rules/design-principles.md` gains two sections. _Meaning never rides on a glyph_ is the surviving third reason from the issue. _Conformance with the toolchain we belong to_ goes beyond what #339 asked for: the console decision was the second unwritten instance of a rule we already follow (verbosity forwarding is the first, `NO_COLOR` a third), and it now also records that conformance is about behavior rather than source, after this PR briefly contained a verbatim copy of an upstream method that brought an attribution obligation and a resync duty with it.
- `.claude/CLAUDE.md`: one index line, updated because the section list above invalidated it.
- `src/Buildvana.Core.ConsoleOutput/ConsoleReporter.cs` gains a sentence of `<remarks>`. It justified reading `Console.Out` and `Console.Error` at every write by `Console.SetOut` / `Console.SetError`, which nothing in this repository calls, while the reason that actually load-bears arrived with this PR and was recorded only on the new class: replacing the console's output encoding flushes both cached writers and drops them without closing them, twice per run, so a captured one keeps writing at an encoding the console has left behind. The coupling is breakable from either side and was described from one.
- `docs/EnvironmentVariables.md` now documents every variable `bv` reads, rather than four out of ten. The page opens by promising exactly that, and this PR is what made the gap self-referential: the section it adds is the first place in the docs to reason about `NO_COLOR`'s convention, on a page where a reader could not then look `NO_COLOR` up. `TERM`, `GITHUB_ACTIONS`, `GITHUB_OUTPUT`, `GITLAB_CI` and `CI_SERVER_HOST` are documented too, in preference to qualifying the promise.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [x] Other — agent instruction files under `.claude`

### Breaking changes

- [ ] Yes
- [x] No

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Thanks — all four addressed, three of them in commits. Gate is green (`bv pack` 0/0, inspectcode 0/0/0).

### 1. `docs/EnvironmentVariables.md` — done, wider than `NO_COLOR` (5a7b845)

The page promised every variable and listed four of ten, so the promise is now kept rather than qualified: `NO_COLOR`, `TERM`, `GITHUB_ACTIONS`, `GITHUB_OUTPUT`, `GITLAB_CI`, `CI_SERVER_HOST`. `NO_COLOR`'s entry states the any-non-empty-value rule _and_ why it differs from `DOTNET_CLI_CONSOLE_USE_DEFAULT_ENCODING`'s literal `1`, which is the argument that had nowhere to land before.

The CI-detection four are a judgement call, as you say. They went in because the alternative was weakening the opening sentence, and because "read to identify the host" is still read. They will want a revisit when the server-adapter platform/git-host split lands.

### 2. Catch list — taken, but not as a fourth `catch` (a8a841f)

The gap is real and it is two types wide, not one. `HandleSetConsoleEncodingError` swallows `ERROR_INVALID_HANDLE` and `ERROR_INVALID_ACCESS` and routes everything else through `Win32Marshal.GetExceptionForWin32Error`, which maps `ERROR_ACCESS_DENIED` to `UnauthorizedAccessException` and `ERROR_OPERATION_ABORTED` to `OperationCanceledException`. Neither was caught.

Adding one more clause would have extended an enumeration, which is what 3802d37 had just finished removing. Instead the whole set now sits behind a named private filter, `IsConsoleEncodingFailure`, so "what does a console refusing to be reconfigured look like?" is answered in one place, with the reasoning for each type next to it.

Two deliberate exclusions, both stated in that comment:

- **`Exception.IsIORelatedException`**, the repo's existing classification, is a strict superset and was the obvious reuse — but it means "related to the I/O operation this block surrounds", and it admits `ArgumentException` to account for malformed paths. No path reaches this class, so an `ArgumentException` here could only mean we handed the console something it cannot accept. That is a bug to fix, not a console to leave alone, and it must not be swallowed.
- **`OperationCanceledException`** is left out on the same grounds: it is not this scope's cancellation to absorb, and an aborted codepage call is not a failure mode anyone will meet.

One thing your review praised is gone as a consequence: both methods now share the filter, so `Restore` no longer documents that it need not handle `NotSupportedException`. That distinction only ever described an unreachable branch, and one classification in one place seemed worth more than the note.

### 3. `Encoding.UTF8` — went past the nit and changed it (0a9da11)

You are right that the live edge is the substantive part, so it is now removed rather than documented. Measured at CP850, same shape as the other runs:

| encoding passed to the setter | `Console.OutputEncoding.GetPreamble()` | a later `new StreamWriter(stream, Console.OutputEncoding)` writes |
|---|---|---|
| `Encoding.UTF8` | `EF BB BF` | `EF BB BF E2 9C 93` |
| `new UTF8Encoding(false)` | empty | `E2 9C 93` |

The character's own encoded bytes are identical, so console output does not change. The comment now records why this is the one point where following the CLI would have been the wrong instinct: the two upstreams disagree (`Encoding.UTF8` vs MSBuild's BOM-less encoding), so there is no single toolchain answer to conform to, and the choice falls to whichever leaves the smaller trap.

The `Console.Out`/`ConsoleReporter` cross-reference you flagged as missing is in the same commit, in `<remarks>`: replacing the output encoding flushes and discards the cached `Console.Out` and `Console.Error`, restoration does it again after the reporter exists, and that is safe only because the reporter reads those properties at every write. Its decision, on which this class depends — worth saying out loud, since a change to either side would break the other silently.

### 4. Changelog section — deliberate, no change

_Changes to existing features_ is right. This is a behavior change, not a fix to something that shipped broken: no released `bv` ever promised a UTF-8 console. The `bug` label has since been removed from both issue and PR, which agrees.

### Description

Both points you flagged as belonging there are now in it: the `ProcessRunner` leg — child `dotnet` output decoded from the pipe at UTF-8 and then best-fit mangled on the way out, because nobody owned the console — and the docs fix under _Additional changes_.

rdeago:
Thanks — all four taken, in five commits. Gate is green (`bv pack` 0/0, 827/827 tests, inspectcode 0/0/0).

Two of them went further than the ask: the coverage finding turned out to be two lines rather than one, and on `OperationCanceledException` I ended up changing the code you said needed no change. Details below.

### 1. Coverage — right, and it was two lines (ef7ed0b, 401cf93)

Your prediction was correct and your mechanism was correct. Measured on the report from the last gate run before these commits:

```
CLASS Buildvana.Tool.Infrastructure.ConsoleEncodingScope   line-rate = 0.333
  IsDefaultEncodingRequested(string)    line 110   hits=1
  IsConsoleEncodingFailure(Exception)   line 178   hits=0
  .cctor()                              line  58   hits=0
```

The second uncovered line is the `Utf8NoBom` field initializer, which compiles to a static constructor. It could not appear in your evidence: the `+1 line, +1 hit` you reasoned from is `bbb6844`, and the field arrived in `0a9da11`, which has not been through Codecov. The class is `beforefieldinit` and `IsDefaultEncodingRequested` touches no static field, so the existing test never triggers initialization.

Both fixed your way, for your reason. `[ExcludeFromCodeCoverage]` was doubly wrong on the predicate — nothing about it is environment-owned, so the justification could only have restated the exclusion — and the same holds for the encoding. Coverage was the symptom; the disease was that the file's two decisions a reader could get wrong *while leaving everything compiling and the console behaving identically* were the two things nothing exercised. Reverting `Utf8NoBom` to `Encoding.UTF8` is exactly that shape of edit: same codepage, same bytes for every character `bv` prints, preamble quietly back on the property.

So both are `internal` and pinned — ten cases for the classification, and same-codepage-no-preamble for the encoding. `line-rate` is now 1.

### 2. `OperationCanceledException` — I went past the note and changed the filter (ef02ea2)

You flagged the rationale and left the outcome alone. I think your own reasoning argues for changing the outcome, so I did, and I'd rather you push back on that than have it slip through.

Verified before touching it: `ConsolePal.Windows` routes every `SetConsoleOutputCP` failure that is not `ERROR_INVALID_HANDLE` or `ERROR_INVALID_ACCESS` through `Win32Marshal.GetExceptionForWin32Error`, which maps `ERROR_OPERATION_ABORTED` to `new OperationCanceledException()` unconditionally — no path check, unlike the `ERROR_ACCESS_DENIED` arm.

You are right that it is not a real cancellation: no token reaches this class, whose guarded calls read one property and write another. But that is precisely what makes it belong *in* the filter rather than merely making its old rationale wrong. What it reports is a console call that did not take effect — the filter's exact subject. And excluding it pinned the opposite choice: let it escape. Your own observation is the sharpest argument against that. The constructor runs before `Program`'s `try` and `Dispose` after its `catch` blocks, so an escape is an unhandled crash over a cosmetic concern; and were it inside that `try`, exit code 130 would report a Ctrl-C nobody pressed. Neither is a defensible outcome in a class whose whole premise is that a console we cannot reconfigure is not a reason to fail the run.

Unreachable in practice either way, as we both concluded, so no observable behavior changes — only which outcome the code would pick if it ever happened. The comment keeps the durable half of the old argument and drops the cancellation-ownership half, as you suggested. The `ArgumentException` exclusion is untouched.

### 3. `GITHUB_ACTIONS` — done (c0e354c)

Half a clause, as you said: the comparison is case-insensitive, so `TRUE` and `True` count as well.

### 4. Cross-reference — done (cbf7421)

`ConsoleReporter`'s `<remarks>` now names what actually load-bears, in the terms you put it: `SetOut`/`SetError` is the obvious reason and nothing in the repo calls them, while replacing the console's output encoding discards both cached writers, twice per run, the second time long after the reporter exists. Recorded as a downstream fact in prose rather than a `<see cref>` — `Buildvana.Core.ConsoleOutput` cannot reference `Buildvana.Tool`, and should not learn to.

Also in the description: the coverage paragraph now names the three pinned decisions rather than one, and `ConsoleReporter.cs` is listed under _Additional changes_, since it is a file this PR touches that #339 never asked for.

rdeago:
All three taken, in three commits — 9c2d0be, 437f3ad, d433e52. Comments only: no code changed, and nothing in the built output differs. Gate green (`bv pack` 0/0, 827/827 tests, inspectcode 0/0/0).

Both premises you flagged as recalled-not-read are now read. One of them has a correction.

### 1. The completeness claim — true, and one item of your enumeration is not in the file (9c2d0be)

I had `Win32Marshal.cs` and `ConsolePal.Windows.cs` on disk from the first round, so I could check rather than re-derive. Your conclusion holds exactly. Your list has one member the table does not: there is **no `DriveNotFoundException` arm**. Harmless — it derives from `IOException`, so its presence or absence changes nothing about coverage — but you asked for the enumeration to be confirmed before it became a comment claiming completeness, and that is the one place it was off.

What the switch actually contains:

| arm | produces |
|---|---|
| `FILE_NOT_FOUND`, `PATH_NOT_FOUND`, `FILENAME_EXCED_RANGE` | `FileNotFoundException`, `DirectoryNotFoundException`, `PathTooLongException` |
| `SHARING_VIOLATION`, `ALREADY_EXISTS`, `FILE_EXISTS`, `INVALID_PARAMETER`, default | `IOException` |
| `ACCESS_DENIED` | `UnauthorizedAccessException` |
| `OPERATION_ABORTED` | `OperationCanceledException` |

Three families, all three in the filter, and no arm produces `ArgumentException`. One detail tightens it further than either of us had: `HandleSetConsoleEncodingError` calls `GetExceptionForWin32Error(lastError)` with **no path**, and the `ALREADY_EXISTS` and `FILE_EXISTS` arms `goto default` when the path is empty. So on this leg specifically, two arms that are otherwise path-dependent collapse to plain `IOException` — the table is narrower here than in general.

Written below the list rather than above it, so it sits against the `ArgumentException` exclusion it supports, which is the part that gains most: that exclusion was an assertion about intent and is now a consequence of the table. Scoped to that leg only, and the comment says so — `SecurityException` and `NotSupportedException` arrive from elsewhere, and Unix has its own `ConsolePal` that never consults it. A claim of totality over *everything these setters can throw* would be a different and much weaker claim.

### 2. Coverage — nothing to do

Agreed on all of it, including which part will still be doing work in a year. Nothing changed here.

### 3. The nit is right, and there were two of them (437f3ad)

Confirmed at `Console.cs:141-150`: `s_out.Flush()` then `Volatile.Write(ref s_out, null!)` — flushed and dropped, never closed, and the stream underneath keeps the same handle. So a captured writer keeps working at its construction encoding, exactly as you say, and "abandoned" pointed at an `ObjectDisposedException` that never comes.

The same error, milder, was in `ConsoleEncodingScope`'s own `<remarks>`: "a writer over a stream that has moved on". Both corrected in the one commit — fixing the flagged half and leaving its twin is how the twin survives. Both now say what the failure actually is: UTF-8 bytes onto a console restored to CP850, which is the best-fit mangling this PR exists to remove, reintroduced by the one class that renders `bv`'s output. The PR description had inherited the same phrasing and has been corrected too.

### 4. The crack is real, unreachable, and now written down (d433e52)

Premise verified: `Console.cs:136` sets the codepage, `143` and `148` flush. So an `IOException` out of that flush would leave the console changed while `ReplaceWithUtf8` returns `null` and `Dispose` restores nothing — the console left in UTF-8 after `bv` exits, precisely as you described.

Slightly worse than you described, in fact: `s_outputEncoding` is written at `152`, *after* both flushes. So on that path the runtime's own cached `Console.OutputEncoding` keeps reporting the old encoding while the console sits at the new one. Upstream's business, not ours, and it only reinforces the conclusion.

Which is yours: it cannot happen, because `Program.cs:39` is the first statement of `Main` — before `AnsiConsole.Console` on line 41, before anything writes — so `s_out` and `s_error` are null and neither `if` runs.

I still wrote it down. Not as a defect, but because the sentence it qualifies (`a field ends up non-null only if this scope actually replaced that encoding`) is exact about spurious restores and silent about missing ones, and only one of those two is guaranteed by anything in that file. The other is guaranteed by a construction site in a different file. Moving that site later — after argument parsing, say, so `--color` could be read first — is an ordinary-looking refactor that would open the hole with nothing in either file to object to it. Four lines, and the last one says the scope stays first in `Main`.

---

Worth another look, narrowly: the completeness sentence in 9c2d0be is the one thing here that could be wrong in a way that matters, since it is now the load-bearing argument for the `ArgumentException` exclusion. If its scoping reads as overclaiming to you, I would rather hear it now than have it stand as a comment that promises more than the table delivers.

rdeago:
Both taken, in two commits — 435cb69 and 1c3e987. Comments only again: no code changed, nothing in the built output differs.

The sub-caveat you flagged as recalled turned out to be the load-bearing one. It is right, and it changes the fix from the one you proposed.

### 1. The flush is a second leg *into* the table, not something outside it (435cb69)

`WindowsConsoleStream.Write` does route through `Win32Marshal.GetExceptionForWin32Error` — `ConsolePal.Windows.cs:1190-1197`, via `WriteFileNative`, and with **no path**, same as `HandleSetConsoleEncodingError`. So your recollection was correct, and the consequence is that your suggested wording would have put the flush on the wrong side of the line: _"along with the writer flush the setter performs"_ appended to a sentence about what arrives from outside the table would have made it an exception to the claim, when it is another instance of it.

It goes further than that. `WriteFileNative` returns `ERROR_SUCCESS` for `ERROR_NO_DATA`, `ERROR_BROKEN_PIPE` and `ERROR_PIPE_NOT_CONNECTED` (`:1276-1282`) — a closing pipe does not throw at all. Which means the comment's existing _"what it does not swallow outright"_ was already true of both legs; only the enumeration of legs was short. So the sentence now names both and keeps one predicate over them:

> Two legs reach it from here — the SetConsoleCP / SetConsoleOutputCP calls, and the flush of the console's cached writers that the output setter performs afterwards, which enters the table by way of WindowsConsoleStream.Write.

### 2. Your "no action" observation was more than that — it made my sentence false

You were being generous. _"It passes no path, which narrows two arms that would otherwise be path-dependent to plain IOException"_ asserts a narrowing **of type**, and there is none: `ALREADY_EXISTS` (`Win32Marshal.cs:47-50`) and `FILE_EXISTS` (`:58-61`) construct an `IOException` on the path branch and `goto default` to a different `IOException` on the empty one. Message and HRESULT constructor, not type.

And it generalizes, which is the part worth having. Every path check in that table — `FILE_NOT_FOUND` (`:33`), `PATH_NOT_FOUND` (`:36`), `ACCESS_DENIED` (`:44`), `FILENAME_EXCED_RANGE` (`:51`), `SHARING_VIOLATION` (`:54`) — selects a message string; the type is decided by the arm alone. So the honest sentence is stronger than the one it replaces, and it is precisely the one that survives your hypothetical future reader who finds a path being passed somewhere:

> No arm's type depends on the path it is handed, so the classification holds however the table is reached.

The "remaining two clauses" phrasing is gone too. They are named now — `SecurityException` and `NotSupportedException` — which is what stops the sentence being read as an enumeration of everything arriving from outside, the reading that produced your finding in the first place.

### 3. "Before anything writes" — right, and the gap is one line (1c3e987)

Confirmed at `Console.cs:188-211`: `Volatile.Read(ref s_out) ?? EnsureInitialized()`, so the property builds its writer on first **read**. Your demonstration is exact — `Program.cs:41` takes `AnsiConsole.Console`, which hands `System.Console.Out` to Spectre and prints nothing.

That is the whole distance between the condition I wrote and the condition that holds, so the comment now states the necessary one and says how narrow the margin is:

> What keeps the crack shut is that nothing has touched Console.Out or Console.Error yet — not that nothing has printed yet. Either property builds its writer on first read, and the next statement of Main hands Console.Out to Spectre without printing a thing, so moving this scope one line down would be enough to open it.

Output-only is confirmed and in: `Console.InputEncoding`'s setter calls `SetConsoleInputEncoding` and then `Volatile.Write(ref s_in, null)` (`:96-103`) — a reader nulled, no writer flushed. The paragraph says so rather than saying "the setter" while sitting above both calls.

Your sharpening of the `s_outputEncoding` point is right as well, and by the mechanism you named: the getter caches (`:112-124`), so `ReplaceWithUtf8` calling `get()` before `set()` is what populates the field with the *old* encoding, and the throw at `:143` never reaches the write at `:152`. Freshly populated with the stale value by our own call, rather than merely unwritten. Upstream's, and not in the comment.

### 4. One thing I found and deliberately left out

`WindowsConsoleStream.Flush()` opens with `if (_handle == IntPtr.Zero) throw Error.GetFileNotOpen()` (`:1199-1203`), which is an `ObjectDisposedException` — outside `Win32Marshal`'s table and, unlike everything else on that path, **not caught by the filter**. `StreamWriter.Flush()` on a closed writer does the same.

It is a real hole, and it is guarded by the same condition as everything else in §3, one notch stricter: reaching it needs something to have *disposed* `Console.Out`, not merely read it. Nothing in `bv` does, and a future refactor that moves the scope past line 41 arms the flush without arming this. I left it out rather than freight the paragraph with a hazard whose trigger nobody performs — happy to add a clause if you think the filter's boundary should be documented where it is drawn rather than only where it bites.

---

Gate green: `bv pack` 0/0, 827/827 tests, inspectcode 0/0/0.

Two things to look at, both narrow: whether §4 belongs in the file, and whether the rewritten completeness sentence still reads as scoped now that it claims something stronger — it went from a true statement about one leg to a claim over two, and the second leg is one neither of us had read until this round.

rdeago:
Taken, in b484d5b. One clause, in the paragraph you nominated.

You are right, and the part that convinces me is not the general principle but what 435cb69 did to the question. Before it, §4 was a hazard note, and "nobody performs the trigger" was a reasonable test to fail. That commit broadened the subject to _what it does not swallow outright_ and named the flush as leg two — which pulled leg two inside a totality claim while I was separately deciding to withhold the one exit from leg two that bypasses the table. Same fact, different job: it stopped being a hazard and became the boundary of a claim the file makes. I applied the old test to the new situation.

Both facts it turns on are read, including the one you flagged as resting on my §4:

- `WindowsConsoleStream.Flush()` does have the guard — `ConsolePal.Windows.cs:1199-1203`, `if (_handle == IntPtr.Zero) throw Error.GetFileNotOpen()`.
- `Error.GetFileNotOpen()` returns `new ObjectDisposedException(null, SR.ObjectDisposed_FileClosed)`, and it lives in `System.Console/src/System/IO/Error.cs` — the same assembly as `WindowsConsoleStream`, not a shared helper that might differ per consumer. Fetched this round rather than recalled, since your finding rests on it and my §4 was itself the only thing supporting it.
- `ConsoleEncodingScopeTests.cs:55` is as you say. `ObjectDisposedException` derives from `InvalidOperationException`, so the assertion was already there and only the reason was missing.

The clause:

> ObjectDisposedException is excluded on the same grounds, and marks the one place the claim above stops: the flush leg leaves that table through WindowsConsoleStream.Flush's zero-handle guard, or through StreamWriter.Flush over a disposed writer. Nothing here disposes Console.Out, so reaching either would again mean a bug rather than a console refusing to be reconfigured.

Two departures from your draft, both small. It names **both** producers rather than only the handle guard, so the claim can be checked the same way `WindowsConsoleStream.Write` can be checked in the sentence above it — and because the `StreamWriter` one is the more likely of the two to be reached, needing no console-specific state at all. And it says the exclusion is on the same grounds *before* saying what the grounds are, so the clause reads as an exclusion first and a boundary second; the paragraph is about what stays out of the filter, and the totality note is the reason this one is worth stating rather than the point of it.

Your placement argument decided the rest. Its own paragraph would have made a hazard of it after all, and the constructor's note stays about ordering.

On your `DriveNotFoundException` and "no action" corrections: noted, and neither needed conceding to me. The second one in particular — I only rejected the observation because you handed it over in a form precise enough to check.

---

Gate green: `bv pack` 0/0, 827/827 tests, inspectcode 0/0/0.

Not requesting another look, since you called this non-blocking and it landed as one clause. Say so if the two departures above change that.

PR #347, opened 2026-08-13
Default to minimal verbosity, and add the missing "Notice" message level

### LINKED ISSUE #346: Default to minimal verbosity, and add the missing "Notice" message level
### Reference version

`2.1.208-preview`; current `main` at commit `2787117`.

### Background and motivation

Two problems, independently real, but coupled tightly enough that fixing either alone makes things worse.

**1. `bv`'s default verbosity does not match `dotnet`'s.** `bv` defaults to `normal`; the build pipeline commands it wraps (`dotnet restore`, `build`, `pack`, `test`) default to `minimal`. Since `bv` forwards its own verbosity verbatim — `DotNetService.RunDotNetAsync` appends `--verbosity={_reporter.Verbosity}` — plain `bv build` produces a markedly noisier MSBuild log than plain `dotnet build`. Buildvana is a component of a .NET toolchain, and where that toolchain has already settled a question of behavior we should settle it the same way rather than on the merits.

**2. The `MessageLevel` ladder is missing a rung, and two reporters already disagree because of it.** `MessageLevel` and `Verbosity` are documented as mapping one-to-one, with visibility computed as `(int)level <= (int)verbosity`. That leaves no level meaning _significant, but not a warning, and visible at minimal verbosity_ — which is a level MSBuild has had all along as `MessageImportance.High`, and which syslog (RFC 5424) calls `Notice`, at severity 5, exactly between Warning and Informational.

The consequence is in the tree today. `TaskLoggingHelperReporter` maps `MessageLevel.Info` onto `MessageImportance.High`, which MSBuild shows at _minimal_ verbosity; `ConsoleReporter` hides `Info` at `Minimal`. Two implementations of one contract disagree about when the same call is visible. The SDK-side reporter is internally consistent — its `Verbosity` getter correspondingly treats "logs High" as `Verbosity.Normal` — but it had no level to put in MSBuild's `High` slot, so it shifted the whole ladder down a notch. Making every implementation of a contract behave identically in respects like this is precisely what `Buildvana.Core.Abstractions` is for.

**Why they are coupled.** `bv release` records what it did to the repository through ~20 `Info` calls — version spec changed, N public API files modified, changelog substituted, N self-referenced files rewritten, N packages pushed. `release` is CI-only, where that log _is_ the audit trail. Flipping the default to `minimal` without a level that survives it would silence exactly the output that most needs to survive; adding the level without flipping the default would leave problem 1 untouched.

### Proposed enhancement

- `bv` defaults to `minimal` verbosity for every command, with no per-command exceptions.
- A new `MessageLevel.Notice` sits between `Warning` and `Info`, is shown from `minimal` verbosity up, and renders uncolored with a `notice:` label — matching the existing `error:` / `warning:` / `info:` line shape.
- Messages recording a _fact_ — something changed, something was decided, something was deliberately skipped — are `Notice`. Messages narrating what the tool is doing _right now_ stay `Info`. `Notice` is not a louder `Info`; it is a quieter `Warning`, and should feel like it costs something.
- At the default verbosity, `bv build` / `restore` / `test` / `pack` produce output comparable to the `dotnet` command underneath, and `bv release` still prints its complete record.
- A given `MessageLevel` becomes visible at the same point whether it is rendered by `bv`'s console or by an MSBuild task.

### Implementation proposals

Four commits.

**1 — Add the level; fix both reporters' mappings.**

- Add `MessageLevel.Notice` between `Warning` and `Info`.
- `IsEnabled` can no longer be `(int)level <= (int)Verbosity`: six levels over five thresholds is not injective. Add `MessageLevelExtensions.MinimumVerbosity()` to `Buildvana.Core.Abstractions` (a switch expression), and redefine `IsEnabled(level)` as `IsVerbosityAtLeast(level.MinimumVerbosity())`.
- Rewrite the XML docs on both enums: they currently advertise the one-to-one mapping and the integer comparison, and `Verbosity.Normal` still claims to be the default. Document the `Notice` vs `Info` criterion on the enum member itself, where call-site authors will read it.
- `ReporterExtensions`: add the `Notice(string)` and `Notice(CompositeFormat, …)` overloads.
- `TextWriterReporter.StyleFor`: `(null, "notice")`.
- `TaskLoggingHelperReporter` — outbound map:

  | `MessageLevel` | MSBuild call | visible from |
  |---|---|---|
  | `Error` / `Warning` | `LogError` / `LogWarning` | always |
  | `Notice` | `LogMessage(High)` | minimal |
  | `Info` | `LogMessage(Normal)` | normal |
  | `Detail` / `Trace` | `LogMessage(Low)` | detailed |

  and the inverse, which is what the `Verbosity` getter computes:

  | loggers accept | MSBuild is at | getter reports |
  |---|---|---|
  | `Low` | detailed **or** diagnostic | `Diagnostic` |
  | `Normal`, not `Low` | normal | `Normal` |
  | `High`, not `Normal` | minimal | `Minimal` |
  | nothing | quiet | `Minimal` (floor) |

  Both judgment calls in that second table over-claim deliberately, and the class remarks should say so: reporting `Detailed` on the top row would make `Report(Trace, format, args)` short-circuit and drop `Trace` even under `-v:diag`, and flooring at `Quiet` would make `Warning`'s formatting overload short-circuit and drop warnings that MSBuild's own quiet verbosity still prints. The remarks should also record that `Detail` and `Trace` share `Low` because MSBuild's ladder has three rungs and ends there — see _Additional information_.
- Activity lines stay gated at `Info` on both sides. `TaskLoggingHelperReporter` already logs them at `Normal` importance specifically so they are hidden at MSBuild's default verbosity, and keeping the console side at `Info` preserves that agreement and keeps `bv build` from narrating what `dotnet build` does not.

**2 — Reclassify the call sites.** Under the criterion above: in `ReleaseCommand`, the outcome and skip-reason lines become `Notice` while "Reading release asset lists…" stays `Info`; `VersionAdvanceCommand`'s "Version spec changed from X to Y" becomes `Notice`; `DotNetService`'s "Pushed N packages to {source}" becomes `Notice`. The `dotnet nuget push` child-output gate stays at `OutputStreaming.AtVerbosity(Verbosity.Normal)` — the raw push output is narration, and the `Notice` line carries the outcome.

For the avoidance of doubt: `bv version show` and `bv update` print their reports directly to `IAnsiConsole`, bypassing the reporter and its verbosity gate entirely. That path is deliberate and stays as it is — those reports are command deliverables rather than leveled messages, and they are not candidates for `Notice`.

**3 — Flip the default.** `Program.cs` falls back to `Verbosity.Minimal` when `--verbosity` is absent, and `CreateDefaultReporter` matches. Remove `defaultVerbosity` from `ImplementsCommandAttribute`, `CommandRegistration`, and `CommandRegistry`, along with its only call site on `version show` — with `Minimal` as the base, the carve-out it exists for disappears. Update the `--verbosity` help text in `GlobalSettings`, which currently says "Defaults to normal" and is already wrong for `version show`. Changelog entry.

**4 — Align the CI workflows with the new default.** In `release.yml`, the `buildvanaVerbosity` choice input carries only three of the five values and defaults to `normal`; give it the full set in ladder order and default it to `minimal`:

```yaml
      buildvanaVerbosity:
        description: 'Buildvana verbosity'
        required: true
        type: choice
        default: 'minimal'
        options:
        - quiet
        - minimal
        - normal
        - detailed
        - diagnostic
```

Drop the `if [ -z "$BV_VERBOSITY" ]` line in the release step — the input is required and defaulted on a dispatch-only workflow, so it can never be empty — keeping the `RUNNER_DEBUG` override. In `build-test-pack.yml`, which has no such input at all and therefore pins `Normal` unconditionally on every push and pull request, stop passing `--verbosity` and let `bv`'s own default apply, preserving the `RUNNER_DEBUG` override:

```yaml
      run: |
        if [ "$RUNNER_DEBUG" = "1" ]; then
          dotnet bv pack --verbosity diagnostic
        else
          dotnet bv pack
        fi
```

Normalize the verbosity literals to lowercase in both files. No changelog entry for this commit: these are Buildvana's own CI workflows, not behavior we ship.

### Usage examples

`bv build`, before — bv narration plus an MSBuild log at `normal`:

```text
[1] Build: starting...
info: Restoring NuGet packages for solution...
  Determining projects to restore...
  All projects are up-to-date for restore.
info: Building solution (restore = False)...
  Buildvana.Runtime -> D:\...\Buildvana.Runtime.dll
  (… every target's output …)
[1] Build: done (34.2s)
```

after — what `dotnet build` would have printed:

```text
  Buildvana.Runtime -> D:\...\Buildvana.Runtime.dll
Build succeeded in 34.2s
```

`bv release`, after — the record survives at the new default:

```text
notice: Version spec changed from 2.1-preview to 2.1
notice: 3 public API files were modified.
notice: Changelog update skipped: release.changelogUpdates is false.
notice: The post-release hook modified 1 file.
notice: Pushed 4 packages to https://api.nuget.org/v3/index.json
```

`bv build -v normal` reproduces the old default output exactly.

### Risks

- **Breaking, by design.** Every command's default output changes. CI logs get quieter; anyone who wants today's behavior passes `-v normal`. No migration is required, nothing fails, but the change is user-visible and gets a `**BREAKING CHANGE**:` entry under _Changes to existing features_.
- **`MessageLevel` gains a member.** `Buildvana.Core.Abstractions` is not packaged and has no `PublicAPI.*.txt` tracking (only `Buildvana.Runtime` does), so there is no public contract impact. In-repo, every `IReporter` implementation needs the new case: `TextWriterReporter`, `TaskLoggingHelperReporter`, `NullReporter`, `CaptureReporter`, `StringWriterReporter`. The `switch` in `TaskLoggingHelperReporter.Report` throws on unknown levels, so a missed one fails loudly.
- **`Notice` creep.** The failure mode is every call-site author deciding their message is significant, until `minimal` is as loud as `normal` is today. Mitigated by putting the criterion in the enum's own XML doc rather than only in this issue.
- **`Trace` still leaks at MSBuild's `-v:detailed`.** Not fixable; see below.
- With `quiet` now offered in `release.yml`, dispatching a release at that level hides the release record entirely — `Notice` is gated at `minimal`. That is the level doing its job, not a defect, but it is worth knowing that `quiet` means "errors only" for an operation that mutates and pushes.
- Tests asserting the current level-to-visibility relationship will need updating.

### Additional information

**Why `Trace` cannot be confined to diagnostic verbosity inside MSBuild.** `EngineServices` exposes only `Version`, `LogsMessagesOfImportance(MessageImportance)`, and `IsTaskInputLoggingEnabled`. There is no verbosity to read, and this is by design: `LogsMessagesOfImportance` is documented as returning true when an importance is "not guaranteed to be ignored by registered loggers" — a union query across loggers, each with its own verbosity, so a single "MSBuild verbosity" does not exist to be asked for. Both detailed and diagnostic log `Low`, making them indistinguishable from inside a task. `IsTaskInputLoggingEnabled` is documented as a task-input double-logging optimization and is independently settable via `MSBUILDLOGTASKINPUTS`, so reading it as a verbosity flag would be using a convention against its owner's terms. `Trace` therefore rides on `Low` alongside `Detail`.

**Alternatives considered and rejected.**

- _An ungated write on `IReporter`_ (a method that bypasses the verbosity gate, mirroring what `ChildOutput(line, null)` already does for child processes). Rejected: it treats the symptom. If a message must be visible at minimal verbosity, that is a statement about its level, and the fix belongs in the ladder.
- _Keeping `normal` as the default for non-pipeline commands._ Rejected: `bv release` runs the build pipeline itself, and the verbosity knob is process-wide, so the carve-out would restore the noisy MSBuild log in the command with the longest log of all — while keeping the per-command default this issue removes.
- _Having `bv` pass its verbosity down as an MSBuild property for the SDK to read._ Rejected: the SDK would then log differently depending on whether `bv` or `dotnet` started the build, and Buildvana SDK is consumed by projects that may never run `bv`.

**Acceptance criteria.**

- [ ] `MessageLevel.Notice` exists between `Warning` and `Info`, visible from `Minimal` up, rendered uncolored with a `notice:` label.
- [ ] `IsEnabled` no longer relies on the enums' integer ordering.
- [ ] `TaskLoggingHelperReporter` and `ConsoleReporter` agree on the verbosity at which each `MessageLevel` becomes visible; the SDK-side over-claims and the `Detail`/`Trace` collapse are documented in the class remarks.
- [ ] The `Notice`-versus-`Info` criterion is documented on the enum member.
- [ ] `bv` defaults to `minimal` for every command; `defaultVerbosity` and its plumbing are gone; the `--verbosity` help text is accurate.
- [ ] `bv release` prints its full record at the default verbosity; `bv build` at the default verbosity is comparable to `dotnet build`.
- [ ] `release.yml` offers all five verbosity values and defaults to `minimal`; `build-test-pack.yml` no longer pins a verbosity; both keep the `RUNNER_DEBUG` → `diagnostic` override.
- [ ] `CHANGELOG.md` is updated under `## Unreleased changes`.
- [ ] `dotnet run .claude/tools/inspect.cs --gate` exits zero.

rdeago:

- [x] Closes #346
- [ ] Partially closes #

## Proposed changes

Two problems, independently real, coupled tightly enough that fixing either alone makes things worse.

**`bv`'s default verbosity did not match `dotnet`'s.** `bv` defaulted to `normal`; the pipeline commands it wraps default to `minimal`, and `bv` forwards its own verbosity to them verbatim — so plain `bv build` produced a markedly noisier MSBuild log than plain `dotnet build`. Where the toolchain we belong to has already settled a question of behavior, we settle it the same way.

**The `MessageLevel` ladder was missing a rung.** There was no level meaning _significant, but not a warning, and visible at minimal verbosity_ — the thing MSBuild has always had as `MessageImportance.High` and syslog calls `Notice`. Its absence was visible in the tree: `TaskLoggingHelperReporter` had nothing to put in MSBuild's `High` slot, so it shifted the whole ladder down a notch and mapped `Info` there, while `ConsoleReporter` hid `Info` at minimal. Two implementations of one contract disagreed about when the same call is visible.

They are coupled because `bv release` records what it did to the repository through some twenty `Info` calls, and on CI that log _is_ the audit trail: flipping the default without a level that survives it would silence exactly the output that most needs to survive.

Four commits:

1. **Add the level; fix both reporters' mappings.** `MessageLevel.Notice` goes between `Warning` and `Info`. Six levels no longer fit five thresholds injectively, so `IsEnabled` can no longer compare the enums' underlying values; the new `MessageLevelExtensions.MinimumVerbosity()` states the mapping and is now the single authority both reporters answer to. `TaskLoggingHelperReporter`'s outbound map shifts by one rung and its `Verbosity` getter with it. Both enums' docs lose the one-to-one claim they can no longer make, and the `Notice`-versus-`Info` criterion goes on the enum member itself, where the next call-site author will meet it.
2. **Reclassify the call sites.** Messages recording a fact — something changed, was decided, or was deliberately skipped — become `Notice`; messages narrating what the tool is doing right now stay `Info`.
3. **Flip the default,** and remove the per-command `defaultVerbosity` plumbing, whose only carve-out disappears with `minimal` as the base.
4. **Align the CI workflows** with the new default.

At the default verbosity, `bv build` / `restore` / `test` / `pack` now produce output comparable to the `dotnet` command underneath, `bv release` still prints its complete record, and a given `MessageLevel` becomes visible at the same point whether `bv`'s console or an MSBuild task renders it. `bv build -v normal` reproduces the old default output.

Two judgment calls in `TaskLoggingHelperReporter` are documented in its remarks rather than left to be rediscovered: the `Verbosity` getter over-claims at both ends of the ladder on purpose (an honest `Detailed` would make the formatting overloads drop `Trace` under `-v:diag`; an honest `Quiet` would make them drop warnings MSBuild still prints), and `Detail` and `Trace` share `MessageImportance.Low` because MSBuild's ladder has three rungs and ends there — `EngineServices` exposes no verbosity to read, by design, so detailed and diagnostic are indistinguishable from inside a task.

## Additional changes

Beyond what the issue laid out:

- **More call sites are promoted to `Notice` than the issue enumerates**: `VersioningService`'s "Version X (height N, publicity)" line, `ServerRelease`'s "Repository unchanged, no commit to push.", `ReleaseCommand`'s committer-identity and empty-changelog-substitution lines, `DotNetService`'s "No test projects found, skipping tests." — which nothing else recorded, leaving `bv test` on a solution with no test project indistinguishable from one whose tests all passed — and all three of `bv version advance`'s outcome lines rather than the one the issue names, the command having no deliverable stream of its own to carry the rest. Each records a fact, which is where the criterion puts it. `VersioningService`'s is also collateral of the remap: it is Core code that runs inside SDK tasks too, where `Info` used to mean `MessageImportance.High`, so leaving it would have removed it from every plain `dotnet build`. Side effect worth knowing: `bv version show` now emits that one line on standard error alongside its report, which is untouched and still pipes clean.
- **`bv release` gains the outcome lines it never had.** The reclassification could only promote messages that already existed, and neither the push, nor the publication, nor the rollback had one: at the new default, a successful release announced everything it had prepared and nothing it had published. `GitService.Push` now records the branch and the remote, `ServerRelease.PublishAsync` the published tag and its asset count, and each rollback step what it undoes — all in the server-independent layer, so a future adapter inherits the record instead of reinventing it. The publication line comes last, past the point where the rollback actions are cleared, so it cannot claim a release that a later failure would delete.
- **`HookRunner` states that a hook ran**, which nothing recorded. The absent-hook message stays at `Info`: its cost scales with the number of hook events a command raises, and a repository with no hooks at all would pay a notice per event to be told about files it never had.
- **The hook args dump drops from `Detail` to `Trace`.** A whole JSON document on one line is diagnostic chatter, and the same document is written to the hook's args file and left there after the run.
- **`GITHUB_OUTPUT` is read before the release exists.** The Actions step output is written after publication, and reading the variable there let one that was never set fail — and therefore roll back — a release that had otherwise succeeded. `GitHubServerRelease` now requires it in its factory method, before the draft release is created, and remembers the path for the life of the release; `SetActionsStepOutput` takes the path from its caller and keeps only the Actions-specific part. `docs/EnvironmentVariables.md` states the new guarantee, and the changelog carries the entry: this changes released behavior.
- **The public API count loses its unreachable singular.** `TransferAllPublicApisToShipped` yields both files of every pair it modifies, so the count is always even and `1 public API file was modified.` could never be printed. The switch is now an expression over the two cases that exist.
- **`TaskLoggingHelperReporter.Verbosity` collapses from three branches to two.** The issue's inverse table has four rows, but its bottom two both yield `Minimal`, which makes the `High` probe dead code. Same semantics.
- **Two unreleased `CHANGELOG.md` entries are corrected in place** rather than left to contradict the new ones in the same release: `bv version show`'s own minimal default, and the resolved `--verbosity` forwarded to `dotnet`.
- **Code this branch did not otherwise change gets tests.** `ReporterExtensions`' whole shortcut-and-formatting family was uncovered before the `Notice` overload joined it, `Report`'s null check and skip-formatting short-circuit included. So was every counted line a release prints except its plural case, in `ReleaseCommand` and in `ServerRelease`, along with the guards that keep a release rollbackable — its one caller never calls a method out of order, always has a commit to push, and adds at most one post-release commit. The command's cases go through `ReleaseHarness`, which grows the one knob they needed: how many of the three self-reference targets the repository has.
- **Lines already over the limits are brought within them**, in the files this branch works in: `ReleaseCommand`, `ServerRelease`, and `GitHubServerAdapter`, plus `ReleaseCommand`'s primary constructor under the stricter 120-character limit on declarations. Each such sweep is its own commit, so the change under review stays readable.
- `MessageLevelExtensions` throws a plain `ArgumentOutOfRangeException` instead of using `CommunityToolkit.Diagnostics`' `ThrowHelper`: `Buildvana.Core.Abstractions` has no package references at all, and a throw helper is not reason enough to add one.

Nothing is done about `GitHubServerAdapter` and `GitHubServerRelease`, whose Octokit calls account for most of what Codecov's patch check still misses: neither file has ever been covered on `main` either. `[ExcludeFromCodeCoverage]` on the methods that do nothing but make those calls is defensible — that is behavior the environment owns — but it moves the badge, and it is a policy decision of its own rather than something to settle at this PR's tail.

## Types of changes

- [x] Bug fix 
- [x] New feature 
- [x] Changes to CI (workflows, bot / app configurations)
- [ ] Other

### Breaking changes

- [x] Yes
- [ ] No

Every command's default output changes. CI logs get quieter; anyone who wants today's behavior passes `-v normal`. No migration is required and nothing fails, but the change is user-visible and gets a `**BREAKING CHANGE**:` entry under _Changes to existing features_.

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Finding 1 was the real one, and it would have shipped. Six commits, one per addressed finding.

### 1. The release log records the no-op but not the act — fixed, one level up

The outcome lines go in the server-independent layer rather than in the GitHub adapter: `ServerRelease.PublishAsync` reports the published tag and its asset count once `DoPublishAsync` returns, and `GitService.Push` reports branch and remote after the push. Putting them in `GitHubServerRelease` would have made the audit trail a GitHub feature that the next adapter has to reinvent. (863f7e6)

The force-push branch is reported too: it runs during rollback, so it is the one push whose omission would leave the trail describing a remote state that was subsequently undone.

`Publishing…` and `Pushing…` stay at `Info` — they narrate what is about to happen in front of a network call, which is not what the reader wants afterwards. `Asset upload skipped: no release assets defined.` stays at `Info` as well: `Published release X with no assets.` already records that fact, so promoting the skip line would say it twice.

### 2. "Deliberately skipped" one level down — two of three

- `DotNetService.cs:162` → `Notice`. Agreed, and it was the strongest of the set: `bv test` on a solution with no test project was indistinguishable from one whose tests all passed. (c159204)
- `HookRunner.cs:101` stays at `Info`. Its cost scales with the number of hook events a command raises, and a repository with no hooks at all would pay a notice per event to be told about files it never had. What was actually missing is the positive: `HookRunner` now records at `Notice` that a hook **ran**, which is the fact worth keeping and which nothing outside `bv release` was recording. (0836f1b)
- `GitHubServerRelease.cs:78` — see above.

### 3. `CHANGELOG.md` — the `bv version show` entry is now false

Fixed, and rephrased around what the command actually guarantees rather than around a verbosity level: the report is alone on standard output, so it stays pipeable whatever the verbosity. (33da375)

### 4. `ArgumentOutOfRangeException.ParamName` regresses to `"this"`

Fixed. The throw moves into a static local function whose parameter really is called `level`, so `nameof` keeps doing the work and no literal has to be kept in sync by hand. Both unknown-level tests asserted the exception type alone — which is exactly why this went unnoticed — and now assert `ParamName` too. (c282277)

### 5. Style — three modified lines exceed 140 characters

Wrapped. (394e6a6) The style guide also asks that a file one is actually working in comes fully within the limits, so a separate commit takes care of the five pre-existing over-long lines and of the primary constructor, which the stricter 120-character limit on declarations catches. (57322ee)

### 6. `release.yml` — `quiet` in the dispatch dropdown

Left in place, by @rdeago's call: it is a knob he rarely turns but wants available. Errors are logged at `quiet` too, and a release run's outcome is independently visible as a tag, two commits, and a published release.

### Additional change, not from the review

The hook args dump drops from `Detail` to `Trace` (df14e81): a whole JSON document on one line is diagnostic chatter, and the same document is written to the hook's args file and left there after the run, so nothing is lost by asking for `--verbosity diagnostic` to see it inline.

---

`dotnet bv pack` and ReSharper at warning severity are both clean. @claude, ready for another look.

rdeago:
Both halves of the new finding taken, both nits taken, plus one change of @rdeago's that the first half made obvious. Five commits.

### The publication is announced before it becomes irreversible — fixed

Right, and it would have shipped: `OnPublishedAsync` is the last step that can still fail, and its failure rolls the release back, deleting the release and the tag the notice had already claimed. The notice moves to the end of `PublishAsync`, past `_rollbackActions.Clear()`, so it prints only once nothing can undo the publication. Ordinary-path output is unchanged — the steps it now follows print nothing of their own. (24c402c)

### The rollback's own acts are invisible at minimal — fixed

Agreed: the argument for the force push covers the release deletion and the commit reset verbatim. `DeleteReleaseAsync` now records the release deletion and the tag deletion separately, each right after its act, as the pushes do; `UndoLastCommit` records the reset. The paths that find no tag to delete stay at `Info`, having changed nothing. (7b05f70)

Wording of the reset line is `Undid the last commit.` rather than the branch-and-SHA form: @rdeago's call, and the clearer line.

The third sub-item is declined. The rollback's force push does not need renaming: at minimal it already reads `Force-pushed 'main' to 'origin'.`, distinct from `Pushed 'main' to 'origin'.`. What `Info` hides is the `Force pushing changes to...` narration, not the outcome — and with the two lines above now printing around it, the rollback is unmistakable.

### Nits

- `HookRunner` — taken, and extended to the line above, which repeated the path just as well. The pair now reads `Running hook release/post-release...` / `Hook release/post-release ran.`. The absent-hook line keeps its path: there, the path is the file one would create. (63dbee9)
- Line sweep — taken, for `ServerRelease.cs` and for `GitHubServerAdapter.cs`, which this round works in as well. Seven lines between them. `GitHubServerRelease.cs`'s single long line is its `CreateAsync` declaration, wrapped by the commit below. (ce86eee)

### Additional change: `GITHUB_OUTPUT` is read before the release exists

The first half's diagnosis pointed at more than the notice's position: `SetActionsStepOutput` read `GITHUB_OUTPUT` at the worst possible moment — after a successful publication, where an unset variable fails and rolls back a release that was correct. Moving the notice hides that; @rdeago asked for the cause to go too. `GitHubServerRelease` now requires the variable in its factory method, before the draft release is created — nothing exists to undo yet, and the message names the variable exactly as before — and remembers the path for the life of the release. `SetActionsStepOutput` takes the path from its caller and keeps only the Actions-specific part. `docs/EnvironmentVariables.md` states the new guarantee. (163f84f)

---

`dotnet bv pack` and ReSharper at warning severity are both clean. @claude, one more look, please.

rdeago:
The changelog entry and both nits taken; the observation left as a recorded decision. Three commits.

### The `GITHUB_OUTPUT` fix has no changelog entry — fixed

Agreed, and it belongs exactly where you put it: an unset variable used to fail — and therefore roll back — a release that had already been published correctly, which makes it a sibling of the three `bv release` entries above it. The entry names the message change too. (9e77a49)

### Observation: the two Actions requirements are checked on either side of the build — left as is

Your own objection decides it, and there is a second: after `163f84f` an unset `GITHUB_OUTPUT` costs a wasted build rather than a rolled-back release, and on a runner the variable is always set — reaching the failure at all means running `bv release` with `GITHUB_ACTIONS=true` off a runner, which is a debugging scenario. A pre-flight virtual would also be one more member to place when `ServerAdapter` is eventually split into its platform and git-host halves. Recorded here so it is a decision, not an accident.

### Nits

- **`Undid the last commit.` printed twice** — fixed, one level up. The duplication belongs to the loop, not to the primitive: the rollback walks back the release commit and everything committed on top of it as a single undoing, and only the caller knows what that set was. `UndoLastCommit` keeps its narration and reports no outcome; the loop records the act once, in the vocabulary the rest of the release log uses — `Undid the release commit.`, or `Undid the release commit and 2 post-release commits.` (ec6f634)
- **`Deleted the previously created release.` names nothing** — fixed, but not from `release.TagName`. Your caveat points the right way and understates the problem: the draft's tag name is not sometimes stale, it is always stale. The draft is created at `ReleaseCommand.cs:103`, before the release commit exists, and that commit always moves the height and with it the version — so the tag the draft carries is never the one `PublishReleaseAsync` publishes under, and the local object is never refreshed either, the `Edit` result being discarded. Naming the post-publish deletion from it would print the wrong version; naming the draft deletion from it would print a version that never existed. The name now comes from the caller — `Deleted release 1.2.3.`, pairing with the publish line — and the draft, having no tag of its own, is reported as the draft it is: `Deleted the provisional draft release.`, matching the `Creating a provisional draft release...` that opened it. (e9477b3)

---

`dotnet bv pack` and ReSharper at warning severity are both clean. @claude, one more look, please.

rdeago:
The nit taken, the note recorded. One commit.

### `where a failure costs nothing` — fixed

You are right, and the overreach is mine twice over: the note you are quoting is the one I wrote in the same comment where the entry claims the opposite. The check runs in `GitHubServerRelease.CreateAsync`, reached from `ReleaseCommand.cs:103`, and `:101` is the verification pass — so an unset variable still costs a clean, a build, and a test run. What it no longer costs is an undoing, nothing having been created or changed at that point, and that is the whole of what the fix bought. The clause now says so; `docs/EnvironmentVariables.md` was already precise and needed nothing. (5664fc2)

### The lost `Edit` response — recorded, not closed

Worth having written down, and the window is real: `Release.Edit` applies server-side, the response is lost, `DoPublishAsync` throws before `PublishAsync` registers `UndoPublishAsync`, and the constructor's rollback deletes a release that had stopped being a draft a moment earlier — leaving GitHub's tag behind and calling the thing a draft in the log.

Leaving it. The dangling tag is `main`'s behavior and no line of this branch touches it; distinguishing the case would mean re-reading the release before deciding what to call it, i.e. a network round-trip on the rollback path, to get the wording right in a window measured in the milliseconds between GitHub committing the edit and failing to say so. The wording is correct for every case that is actually reachable, which is the standard the rest of these notices are held to. Recorded here so the next reader of that code finds the window described rather than having to rediscover it.

---

`dotnet bv pack` and ReSharper at warning severity are both clean.

rdeago:
Codecov's patch check, not a review finding. Four commits, and one of them touches production code — flagging that here, since it lands after the ready-to-merge verdict.

### What the red number was made of

`codecov/patch` was red at 60.55%; `codecov/project` passed, +0.08%. Of the 41 missed lines, **31 already existed on `main`, already uncovered**: `Info` → `Notice` and the line-length rewraps are what made git call them new. One `ThrowHelper.ThrowInvalidOperationException(…)` wrapped across two lines is four old misses becoming eight; the `GenerateReleaseNotes` call is one becoming four. Ten lines were genuinely new, seven of them in the two Octokit files.

### The public API count could never say "one" (77e9318)

`TransferAllPublicApisToShipped` yields both files of every pair it modifies — the unshipped file is emptied into the shipped one, and the release commit carries both — so the count is always even, and `1 public API file was modified.` was unreachable. The switch is now an expression over the two cases that exist.

### The counted lines are tested (888654e, c91601c, ae10864)

What was genuinely untested is one thing throughout: every counted line except its plural case. The command's cases go through `ReleaseHarness`, which grows the knob they needed — how many of the three self-reference targets the repository has — and a `Notices` view to assert on. `ServerRelease`'s go through the class directly: it is a contract for every adapter, and its one caller today never calls a method out of order, always has a commit to push, and adds at most one post-release commit. So the guards that keep a release rollbackable, and the undone-commit and asset counts, are asserted by driving it over the harness's repository.

Measured on the Release build the gate produces: `ReleaseCommand.cs` 146/158 → 157/157, `ServerRelease.cs` 118/162 → 132/162, every one of the 19 patch misses now hit and both partials resolved.

### The Octokit files are left alone

`GitHubServerAdapter` and `GitHubServerRelease` contribute 18 of the missed lines and have never been covered on `main` either. `[ExcludeFromCodeCoverage]` on the methods that do nothing but call Octokit is defensible — that is behavior the environment owns — but it moves the badge, and class-wide would be too coarse: the adapter also parses the origin URL and composes file URLs. A decision of its own, not one for this PR's tail.

---

`dotnet bv pack` and ReSharper at warning severity are both clean.

PR #348, opened 2026-08-14
Fix the release workflow's hook failure, and the home-directory bug behind it

rdeago:

No issue: this fixes a failing release workflow, and went straight to a PR.

- [ ] Closes #
- [ ] Partially closes #

## Proposed changes

The release workflow failed: the `release/post-release` hook died with
`FileNotFoundException: '/home/runner/work/Buildvana/Buildvana/buildvana.jsonc'`
after the configuration file moved into `.buildvana/`. The hook read the file by
name, relative to its working directory.

That literal was the presenting symptom. The cause underneath it is that
**a configuration file inside `.buildvana/` makes `.buildvana/` itself a home
directory**: the marker rule counted a configuration file "directly in the
directory", and `.buildvana/buildvana.jsonc` satisfies that clause for
`.buildvana` itself. Hooks are projects living in there, so every hook build
walked up, stopped at `.buildvana/`, and took it for the repository. Measured
before the fix:

    HomeDirectory:    D:\...\Buildvana\.buildvana\
    StyleCopJsonPath: (empty)

and after:

    HomeDirectory:    D:\...\Buildvana\
    StyleCopJsonPath: D:\...\Buildvana\stylecop.json

That second line is the `SA1633` warning in the same release log: the repository's
`stylecop.json` was found and then discarded as "outside the repository", so hooks
compiled under StyleCop's defaults instead of ours. Every other
`HomeDirectory`-relative path in a hook build was wrong by one level too. No
analyzer rule was touched to fix it.

**What changed**

- **`.buildvana/` is no longer a configuration-file location.** The file lives in
  the home directory; `.buildvana/` keeps the hooks. The marker rule collapses to
  one sentence with no exceptions — _every marker sits in the directory it marks_ —
  in the loader, in `HomeDirectoryDiscovery`, and in `Sdk.props`, which now probe
  two candidates instead of four. Removing the location makes the collision
  unrepresentable rather than guarding against it.
- **`bv` resolves the configuration file once per run.** `BuildvanaConfigLoader`
  becomes `BuildvanaConfigProvider`, an instance holding both the path and the
  parsed configuration, each resolved on first read, as `HomeDirectoryProvider`
  does for the home directory. The probe is private, so nothing can ask a second
  time and get a different answer. Previously three places found the file
  independently and one of them threw the path away.
- **Hooks are told which file it is.** `RuntimeInfo.ConfigFile` carries the path,
  and `HookArgs.LoadConfig()` reads that same file through the new
  `BuildvanaConfig.LoadFile`, so a hook after settings does not search either —
  which file comes from the args, what it says is read from disk at the moment of
  the call.
- **The post-release hook stops hardcoding the file name.**

**Known limitation, deliberate:** the hook uses `BuildvanaConfig.FindFile()` rather
than `RuntimeInfo.ConfigFile`, with a comment saying what to replace and when. The
SDK pins `Buildvana.Runtime` to its own version, so a hook compiles against the last
published release; using the new member now would fail the hook's compilation during
the very release that publishes it, aborting that release. It becomes a one-line
change in the release after next.

## Additional changes

- **The configuration model owns its defaults.** `dotnet.configuration` → `Release`
  and `versioning.assemblyVersionPrecision` → `Major` used to be resolved by
  whoever read them, so a hook asking the same question of the same file got `null`
  where `bv` got a value — one file describing two different builds. Each default is
  now a constant next to its setting, with an `Effective…` accessor over a nullable
  receiver, so an absent section and an absent setting resolve identically and no
  consumer writes its own fallback.
- **Hooks.md's contract-evolution rule was wrong and is rewritten.** It applied
  binary-compatibility reasoning to a contract with no version boundary: a hook is
  compiled from source at every run against the Runtime version the SDK pins, which
  `bv` refuses to mismatch, and the args file is rewritten immediately before the
  run. What has to stay stable is the source surface, which is why `ConfigFile`
  could be added as a required member.

## Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Other

### Breaking changes

- [ ] Yes
- [x] No

The `.buildvana/` configuration location shipped in exactly one preview
(`2.1.208-preview`) and never in a stable release — its own bullet was still under
`## Unreleased changes` — so its removal edits that bullet rather than adding a
breaking-change entry. Migration for a preview adopter is `git mv`.

## Checklist

- **For code changes only:**


rdeago:

rdeago:
Addressed, one commit per finding. Gate (`bv pack` + ReSharper at WARNING and above) is green: 0 errors, 0 warnings, 0 info.

## 1. `ConfigFile` remarks — done (d7c9ace)

Right, and the contradiction was load-bearing in the worst place. The first `<para>` now says a settings-reading hook gets there through `HookArgs.LoadConfig`; the rest is unchanged.

## 2. Regression tests — finding accepted, fixtures corrected (852fcae)

The finding is right: nothing failed if the location came back. Two of the three proposed fixtures don't demonstrate it, though.

Under the rule as it now stands a configuration file marks the directory it sits in, and `.buildvana` is not special — so `root/.buildvana/buildvana.jsonc` marks `root/.buildvana`, before and after the fix alike. Discovery starting at `root/.buildvana/hooks/release` therefore answers `root/.buildvana` in both, and in `Sdk.props` `_HomeDir_ByConfigJsonc` resolves to `root/.buildvana`, deeper than `.git`, so it wins. Those two tests would pass on the buggy code and fail on the fixed code.

That outcome is deliberate. A repository that keeps its configuration in `.buildvana/` after this change is broken, and the fix for it is the `git mv` this branch opens with, not a guard. It is why the migration diagnostic and the name guard were dropped from the original plan.

What the removed clause actually did was make a directory a home directory because a `.buildvana` subdirectory _of it_ held a configuration file. Hence the fixture:

| fixture | old rule | current rule |
|---|---|---|
| `.git/HEAD` at `root`, config at `root/nested/.buildvana/buildvana.jsonc`, start at `root/nested` | `root/nested` | `root` |

One test per implementation on that fixture (`HomeDirectoryDiscoveryTests`, `SdkPropsTests`), plus `FindFile(dir)` returning `null` for a file left in `.buildvana/`, which is the probe itself (`BuildvanaConfigLoadTests`). Hermetic: the `.git/HEAD` marker is inside the fixture, and any stray marker above the temp directory has a shorter path, so nearest-wins keeps it out.

Each was checked against the old behavior by restoring the probes in all three implementations — `HomeDirectoryDiscovery`, `BuildvanaConfig.FindFile`, and `Sdk.props`: one failure per test project, each of them the new test.

Not added: the fourth test, in `BuildvanaConfigProviderTests`. The provider has no probe of its own — it delegates to `BuildvanaConfig.FindFile` — so a restored probe is caught by the `Buildvana.Runtime` test, and a copy there would pin nothing further.

## 3. Null round-trip — done (b4fd520)

`Load_RoundTripsWhatBvWrites` now runs both ways, off an optional parameter on `SampleArgs` that also spares `LoadConfig_WithoutConfigFile_ReturnsEmptyConfig` its two-level `with` expression.

Both regressions were checked:

- `DefaultIgnoreCondition = WhenWritingNull` on the context — 3 failures, one of them the new null case. (The other two are `ReleaseInfo.PreviousVersion`, required and nullable for the same reason.)
- `[JsonIgnore(Condition = WhenWritingNull)]` on the member — 1 failure, the null case. That is the one only this test catches.

## 4. Hook's interim `FindFile()` — verified, second half done (7faa374)

`FindFile` is public in 2.1.208-preview: `git show 2.1.208-preview:src/Buildvana.Runtime/BuildvanaConfig-Load.cs` has `public static string? FindFile(string? homeDirectory = null)`, and the tag's `PublicAPI.Unshipped.txt` carries the entry. `dotnet build .buildvana/hooks/release/post-release.cs` compiles clean against the pin. No aborted release waiting there.

The argument point is taken: it now passes `hookArgs.RuntimeInfo.HomeDirectory`, so the hook depends on what `bv` said rather than on process state, and the deferred edit is one expression.

## 5. Nits — all three done

- baa2528 — the four entries are sorted; the whole file now matches its sorted order line for line, so the next diff has nothing incidental in it.
- 19e00d1 — `VersioningSettings` cites `VersioningConfig.DefaultAssemblyVersionPrecision`. **And `DotNetSettings.cs:33`, which spelled out `"Release"`** — same drift, same commit, since that is exactly what the defaults commit set out to remove.
- ca00901 — `RuntimeInfo.ConfigFile` in the C# paragraph.

## 6. Observation

Agreed on both counts, and no change made: the alternative puts the fallback back at the call sites, which is the thing that commit removes. Noted as a convention.

---

Nothing here changed the PR description: all of it fixes code this PR introduced, which is not what the "Additional changes" section is for.

@claude please review again.

PR #349, opened 2026-08-15
Stop reporting the version from a constructor

rdeago:
### Description

`bv release` printed the same version line up to twenty-two times per run — four in a row at the head of each pipeline stage, plus two from `bv` itself.

The cause is that `VersioningService` reported the computed version from its constructor. That made the message count a function of how many instances got built, which in the SDK means one per MSBuild node, since the `ComputeVersion` cache is per process. Four lines on a 4-vCPU CI runner; twenty-nine on a 32-core desktop. Nothing about the repository determines it.

The class had a second problem behind the first: it was named a service but behaved as a snapshot. It computed everything in its constructor, `VersionService.Update()` discarded the instance and built another, and the SDK task copied its ten properties into an identically shaped record.

So it is now the pair its own project already establishes with `GitHeightCalculator` → `GitHeightResult`:

- `VersionInfo` — the immutable snapshot, deriving the five version strings from the spec, the height, and the settings. No I/O, and an `internal` constructor: `VersionCalculator` is the only thing that builds one, so the one place that has to get the parameters consistent is the only place that can pass them.
- `VersionCalculator` — the lookups (version file, prerelease tag validation, height walk), returning a `VersionInfo`. Stateless, so calculating again after a commit is the explicit way to get a second answer.

`IReporter` is out of `Buildvana.Core.Versioning` altogether. `CachedVersion` collapses to `(Fingerprint, VersionInfo)` and stays inside the caching partial. Reporting now happens in exactly two deliberate places:

- the SDK task, at **detail** level (`MessageImportance.Low`), invisible below `-v detailed`, where per-node repetition is a diagnostic rather than noise — and where `ComputeVersionTests` reads it as evidence of a cache hit;
- `bv release`, once, just after the tag check — the last thing that can refuse this particular version — where naming it records a decision rather than a guess.

`bv build`/`pack` say nothing about the version, as before.

One consequence worth stating plainly: a release that fails before that point no longer names a version at the default verbosity. What it used to name was a guess — `VersionService` is resolved at the top of the command, before the version spec change and before the release commit, so the old line reported a height at least one below the version that would be published, and named a different version line entirely on a `--bump` run. `-v detailed` still carries the SDK task's line for anyone who wants the provisional number.

### Additional changes

None.

### Checklist

- [x] Build is clean: 0 errors, 0 warnings
- [x] Sanity check green (`bv pack` + ReSharper at WARNING and above, both silent)
- [x] Tests pass: Core.Versioning 102, Sdk.Tasks 41, Tool 399 — including `Release_RecordsTheVersionItPublishes`, which asserts the line appears exactly once
- [x] No changelog entry — the versioning feature is still under _Unreleased changes_ and never documented the line

### DIFF
--- a/src/Buildvana.Core.Versioning/VersionCalculator.cs
+++ b/src/Buildvana.Core.Versioning/VersionCalculator.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (C) Tenacom and Contributors. Licensed under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System;
+using Buildvana.Core.HomeDirectory;
+using CommunityToolkit.Diagnostics;
+using NuGet.Versioning;
+
+namespace Buildvana.Core.Versioning;
+
+/// <summary>
+/// Computes the version being built from the repository's version file, Git height, and versioning settings.
+/// </summary>
+/// <remarks>
+/// <para>The version file (<see cref="VersionFile.FileName"/>, in the home directory) holds a
+/// <c>MAJOR.MINOR[-[tag]]</c> specification; the patch number is the Git height of the version line
+/// (see <see cref="GitHeightCalculator"/>).</para>
+/// <para>An instance carries no computed state: it reads the repository afresh on every
+/// <see cref="Calculate"/> call, so a consumer that changes what the version depends on — by committing,
+/// or by rewriting the version file — calls it again for a <see cref="VersionInfo"/> matching the new state.</para>
+/// </remarks>
+public sealed class VersionCalculator
+{
+    private readonly IHomeDirectoryProvider _home;
+    private readonly VersioningSettings _settings;
+    private readonly GitHeightCalculator _heightCalculator;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="VersionCalculator"/> class.
+    /// </summary>
+    /// <param name="home">The home directory provider used to locate the version file and the Git repository.</param>
+    /// <param name="settings">The versioning settings.</param>
+    /// <param name="heightCalculator">The calculator providing the Git height of the version line.</param>
+    public VersionCalculator(IHomeDirectoryProvider home, VersioningSettings settings, GitHeightCalculator heightCalculator)
+    {
+        Guard.IsNotNull(home);
+        Guard.IsNotNull(settings);
+        Guard.IsNotNull(heightCalculator);
+        _home = home;
+        _settings = settings;
+        _heightCalculator = heightCalculator;
+    }
+
+    /// <summary>
+    /// Computes the version being built, as the repository stands now.
+    /// </summary>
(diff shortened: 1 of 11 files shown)


rdeago:

rdeago:
Addressed. Seven commits; the gate (`bv pack` + ReSharper at WARNING and above) is green on the final state.

**1 — the regression test now tests the regression.** `7db8c84`, taken as written. The bug was twenty-two lines and `Contains` cannot tell one from twenty-two.

```csharp
await Assert.That(harness.Notices.Count(x => x.StartsWith("Releasing version", StringComparison.Ordinal))).IsEqualTo(1);
```

**2 — fixed, though not with the guard.** `3d0c6c6`, `b1cc161`.

The invalid output is real, but a guard throws on a state that a smaller change makes unrepresentable. `ComputeInformationalVersion` chose the separator from `spec.Prerelease`, when the question it actually asks is whether the string it appends to already carries a prerelease part — and that is decided two lines above, by `prereleaseTag`. Reading the answer off the tag makes both of your disagreement cases produce valid output instead of an exception, and leaves behind no defensive branch that no test can honestly reach.

@rdeago's point closes it from the other side: the constructor is now `internal`. `VersionCalculator.Calculate` is the only thing that builds a `VersionInfo`, and it lives in the same assembly, so the one place that has to get the pairing right is now the only place that can pass it at all.

One correction to the premise, since it is what made narrowing free: `Buildvana.Core.Versioning` is a Core-tier project and is not packaged, and it tracks no public API file. "Public and reachable from three assemblies" means our own three.

**3 — no code change; now stated in the description.** The observation is right, but what was lost was a guess rather than information. `VersionService` is resolved at `ReleaseCommand.cs:50`, before the version spec change (`:94`) and before `EnsureReleaseCommit` (`:195`), so the old line reported a height at least one below what would be published, and named a different version line entirely on a `--bump` run. Restoring an early line means announcing the version twice, which is the bug this PR is about. Disclosure it does deserve, so the description now says that a release failing before the tag check names no version at default verbosity, and points at `-v detailed`.

**4 — taken.** `2632cfe`. Announcing `Releasing version 2.3.1.` and then refusing to release it was silly. Below the tag check the comment's claim holds: what can still fail from there is a build, not a verdict on the version.

**5 — taken.** `6e8e468`. `GetOrComputeVersion` returns `VersionInfo`, and `CachedVersion` no longer leaves the file that caches.

Two more, neither of them yours:

- `c4701d9` — Codecov's single partial branch was the `?? string.Empty` at the MSBuild boundary: nothing in the tree ran the calculation against a repository with no commits. Now something does, which also pins the empty-string convention you flagged as correctly placed.
- `f9b1ce5` — three lines in `ComputeVersionTests` ran past 140 characters, two of them lines this branch had changed. The repeated `Count(...Contains(...))` that made them long is a named helper now, and the configuration document moved into a local, since as an argument a multi-line raw literal trips SA1118.

@claude, another look when you have a moment.

PR #351, opened 2026-08-15
Split configuration into wire models and a resolved domain model

### LINKED ISSUE #350: Split configuration into wire models and a resolved domain model
### Reference version

`2.1.259-preview` (last published); commit `348771a`.

### Background and motivation

`BuildvanaConfig` and its section records model the **configuration file**. Nothing models the **settings**. The file is one source among several — the command line is another, and for the committer identity the repository's Git config and the CI server adapter are two more. The settings are the resolution of all of them. Because that second object does not exist, "what is the effective value of X?" has no home, and every consumer builds a private one.

The resolution logic is spread across five different idioms:

| Idiom | Example |
| --- | --- |
| Extension property on the model | `DotNetConfigExtensions.EffectiveConfiguration` |
| Constructor of a settings class | `VersioningSettings`, `DotNetSettings` |
| `Resolve*()` method | `ReleaseSettings`, `VersionAdvanceSettings` |
| Private static helper | `DotNetInvocationsSettings.Resolve` |
| Inline ternary at point of use | `GitHubServerAdapter` ctor, for `github.tokenEnv` |

Which idiom a setting gets is decided by whether it has a command-line flag. A flagless setting can be resolved on the model, so `dotnet.configuration` and `versioning.assemblyVersionPrecision` have `Effective*` properties with their defaults on the record — the right answer. A setting with a flag cannot, because `Buildvana.Runtime` must not know the command line, so its resolution migrates into `Buildvana.Tool` and its default goes with it, out of the packaged contract. The packaging boundary ends up deciding where a default lives on the basis of a fact that has nothing to do with packaging.

Concrete consequences today:

- **The build configuration is resolved four ways**: `DotNetSettings.Configuration`; `ReleaseSettings.ResolveConfiguration()`; a third `??` in `BuildPipeline.RunAsync`; and `bv pack -- -c Debug`, where the forwarded flag reaches `dotnet` after bv's own `-p:Configuration=` and decides the actual build while bv's own resolved value stays `Release`.
- **`release.checkPublicApi` is resolved twice**, by identical three-term chains in `ReleaseSettings` and `VersionAdvanceSettings`, each with its own duplicated `[Description]`.
- **"Unset" has three meanings**: plain `??` nearly everywhere, `is { Length: > 0 }` for `github.tokenEnv`, `IsNullOrWhiteSpace` for `release.emptyChangelog`.
- **`null` is overloaded.** For `dotnet.configuration` it means "unspecified, fall through to `Release`". For `versioning.prereleaseTag` it means "specified as nothing, and that _is_ the answer: prereleases are forbidden". Same declaration shape, opposite semantics, distinguished only by prose in an XML comment.
- **`git.identity` has no consumer at all.** It is schema-validated, publicly exposed, documented — and dead, because the identity chain lives entirely outside the config model.
- **Hooks cannot resolve anything.** A hook receives `BuildvanaConfig` and can correctly answer exactly the two questions that have `Effective*` properties; for everything else it would have to reimplement bv's precedence, which it cannot, because it cannot see the flags bv was invoked with. `RuntimeInfo.ArtifactsDirectory` is the proof: `artifacts/<configuration>` is not derivable from the file, so it is hand-carried across the boundary pre-resolved. Every future setting a hook needs would require the same hand-carry.

**Why now.** This is the last structural change before the 2.1 line goes stable. `Buildvana.Runtime`'s public surface is additive-only once stable, and after this refactor the domain model is what hooks bind against — so it is the thing that guarantee will apply to. Getting it right before the first stable release is the point.

### Proposed enhancement

Buildvana grows an explicit distinction between **wire models** (DTOs faithful to one source's format) and a **domain model** (the single, immutable, fully-resolved answer).

- `Buildvana.Runtime` contains the domain model and nothing else configuration-related. Every default lives on it. It has no knowledge of JSON files, schemas, or command lines.
- `Buildvana.Core.Configuration` contains every wire model — the JSON file's shape and the command-line overrides' shape — plus the one factory that composes them into a domain model.
- Resolution happens exactly once per run, in that factory. No consumer applies a default or a precedence rule.
- Hooks receive the resolved domain model as part of their args, and read settings by property access.
- `git.identity` becomes live.
- The domain model stores environment variable **names** for credentials, and resolves their values on demand through extension properties.

Users see no change to `buildvana.json` / `buildvana.jsonc`, nor to the generated schema's shape.

#### 1. Assembly and type layout

**`Buildvana.Runtime`** — domain model, immutable records, current names kept:

`BuildvanaConfig`, `ReleaseConfig`, `VersioningConfig`, `DotNetConfig`, `DotNetInvocationConfig`, `NuGetConfig`, `NuGetFeedsConfig`, `NuGetFeedConfig`, `GitHubConfig`, `GitConfig`, `GitIdentityConfig`. Enums `ChangelogUpdates` and `AssemblyVersionPrecision` stay where they are.

Removed from `Buildvana.Runtime`: `BuildvanaConfig.Schema`, `JsonFileName`, `JsoncFileName`, `FindFile`, `Load`, `LoadFile` (the whole `BuildvanaConfig-Load.cs` partial); `DotNetConfigExtensions` and `VersioningConfigExtensions` (subsumed by domain defaults); `HookArgs.LoadConfig()`. `BuildvanaRuntimeException` stays — `HookArgs.Load` still uses it, and the new credential accessors throw it.

Added to `Buildvana.Runtime`: `GitHubConfigExtensions` and `NuGetFeedConfigExtensions` (see §7).

**`Buildvana.Core.Configuration`** — wire models, suffixed `*JsonConfig`: `BuildvanaJsonConfig`, `ReleaseJsonConfig`, `VersioningJsonConfig`, `DotNetJsonConfig`, `DotNetInvocationJsonConfig`, `NuGetJsonConfig`, `NuGetFeedsJsonConfig`, `NuGetFeedJsonConfig`, `GitHubJsonConfig`, `GitJsonConfig`, `GitIdentityJsonConfig`. All members nullable; `$schema` lives here. Plus `CommandLineOverrides`, `BuildvanaConfigFactory`, and the file-discovery and loading code moved out of `Buildvana.Runtime`.

Existing types in this assembly are renamed to match: `BuildvanaConfigProvider` → `BuildvanaJsonConfigProvider`, `BuildvanaConfigSerialization` → `BuildvanaJsonConfigSerialization`, `BuildvanaConfigSchema` → `BuildvanaJsonConfigSchema`, and `BuildvanaConfigProviderTests` accordingly. Call sites to update: DI registration, `SelfVersionService` (constructor parameter and its static `LoadFile` call), `HookArgsFactory<TArgs>`, `ComputeVersion.ComputeCore`.

`Buildvana.Core.HomeDirectory` takes its config file names from `Buildvana.Core.Configuration` instead of `Buildvana.Runtime` (Core-to-Core dependency, permitted). `Sdk.props` still mirrors the discovery algorithm; its "apply any change here" note must survive the move.

#### 2. Defaults live on the domain records

As property initializers, never in the factory, and never `required`:

| Domain member | Default | Meaning of the default |
| --- | --- | --- |
| `DotNet.Configuration` | `"Release"` | |
| `DotNet.{All,Restore,Build,Test,Pack,NugetPush}` | empty invocation settings | non-nullable, always present |
| `Versioning.AssemblyVersionPrecision` | `Major` | |
| `Versioning.PrereleaseTag` | `null` | domain option: prereleases not allowed |
| `Release.Branches` | `[]` | no branch produces a public release |
| `Release.Configuration` | — | computed by the factory; falls back to `DotNet.Configuration` |
| `Release.CheckPublicApi` | `true` | |
| `Release.ChangelogUpdates` | `Stable` | |
| `Release.EmptyChangelog` | `null` | domain option: an empty changelog fails the release |
| `Release.Dogfood` | `true` | |
| `GitHub.TokenEnv` | `"GITHUB_TOKEN"` | the _name_; the value is resolved on demand |
| `NuGet.Feeds.{Prerelease,Release}` | `null` | domain option: channel not configured |
| `Git.Identity` | `null` | domain option: no configured identity |

Initializers rather than `required` members is what keeps adding a setting additive once `Buildvana.Runtime` is stable.

**Nullability rule, to be stated in the code:** in a wire model, `null` means "not stated". In the domain model, `null` is a domain option with exactly one meaning. The blank-string normalizations (`emptyChangelog`, `tokenEnv`) collapse into the factory.

#### 3. The factory

```csharp
public static class BuildvanaConfigFactory
{
    public static BuildvanaConfig Create(BuildvanaJsonConfig? json, CommandLineOverrides? commandLine);
}
```

- Contains **no literal default values** — it starts from the domain defaults and overlays.
- Split into one private composer per section (`ComposeRelease`, `ComposeDotNet`, …), so that a single method does not accumulate the whole composition.
- Takes each wire model on its own terms, with no common overlay type. A uniform overlay shape was considered and rejected: a source's vocabulary can include _operations_ — a hypothetical `--no-dotnet-env` clearing every `env` map — which a partial mirror of the domain cannot express.

Merge semantics are **not** uniform:

| Member | Rule |
| --- | --- |
| Scalars | command line beats JSON beats domain default |
| `dotnet.*.args` | append, in order: JSON `all` → JSON per-command → args forwarded after `--` |
| `dotnet.*.env` | layer by key: JSON `all` → JSON per-command (no command-line source today) |
| `release.configuration` | flag → `release.configuration` → resolved `dotnet.configuration` |

Note the args ordering: forwarded arguments must win over configured ones, so they are a _last_ tier and cannot simply be folded into `dotnet.all.args`, which composes first.

#### 4. Two producers, one factory

`bv` composes with a real `CommandLineOverrides`. `ComputeVersion` runs in an MSBuild task host with no command line and passes `null` — "no command line" and "no configuration overrides on the command line" are the same result, so both go through the same factory. `Buildvana.Core.Configuration` already ships inside the `Buildvana.Sdk` package, so this adds no packaging work.

`BuildvanaJsonConfigProvider` keeps its current job — find the file, parse it, validate it against the schema, cache the result — but now yields a `BuildvanaJsonConfig` and the path. Composition happens above it: in DI registration for `bv`, in `ComputeVersion.ComputeCore` for the SDK.

#### 5. The `DotNetService` seam

The factory resolves, per `dotnet` command, the final args and environment (configured tiers folded with forwarded arguments). `DotNetService` still prepends its base arguments (`"build"`, solution path, `-p:Configuration=`) and appends its trailing ones (`ContinuousIntegrationBuild`), because those depend on runtime facts the domain model must not know. `MergeInvocation` shrinks accordingly.

This is also what removes the fourth resolution path for the build configuration: forwarded arguments stop being a separate concept and become a command-line contribution to the same resolution.

#### 6. Hooks

- `RuntimeInfo` gains a required `Configuration` property carrying the domain model.
- `RuntimeInfo.ConfigFile` **keeps meaning "the source file bv read"** — `.buildvana/hooks/release/post-release.cs` rewrites that file as text and depends on it. Its XML doc must say so explicitly, since it loses its "where to load configuration from" role.
- `HookArgs.LoadConfig()` is removed. It has no consumers today.
- Configuration is immutable for the duration of a run; the factory's output is the single source of truth. A hook that rewrites the file mid-run no longer affects what a later hook sees — that was a bug, not a feature.
- Embedding the domain model in each hook's args file (rather than in one shared snapshot) is deliberate: args files are left in place so a hook can be re-run by hand against the last run's args, and a shared snapshot would be overwritten by the next run and desync from a stale args file.

`HookArgs` and `RuntimeInfo` themselves are **not** split into wire and domain forms. Their JSON is unidirectional IPC — no schema, no defaulting, no overrides, and it could be any format. Only the domain form exists.

#### 7. Credentials: names in the model, values on demand

The domain model stores environment variable **names**. It never stores their values, because it is serialized into `.buildvana-temp/hook-args/**` on every run and storing a resolved secret would write the NuGet API key and the GitHub token to disk. **Config-in-`RuntimeInfo` is only safe because credentials are stored as names**, and that coupling must be recorded in the code so it is not later "simplified".

Consumers nonetheless must not have to handle environment variables themselves, so `Buildvana.Runtime` provides extension properties that resolve on access:

```csharp
public static class GitHubConfigExtensions
{
    extension(GitHubConfig @this)
    {
        public string Token => /* reads the variable named by @this.TokenEnv */;
    }
}
```

and likewise `NuGetFeedConfigExtensions.ApiKey` over `ApiKeyEnv`. Extension properties are exactly the right mechanism here: they are not members, so they cannot be serialized — the guarantee is structural rather than a rule someone has to remember. And the model keeps stating, in the member name, that what it holds is a variable name.

They throw `BuildvanaRuntimeException` when the named variable is unset or empty (`Buildvana.Runtime` cannot see `BuildFailedException`); `bv`'s two call sites translate, as `BuildvanaJsonConfigProvider.FindFile` already does for discovery errors. `EnvVarHelper.Require` remains in `Buildvana.Tool` for `GITHUB_OUTPUT`, which is not configuration.

Extension blocks must follow the repository template, including the `CA1034`/`CA1708` suppressions and the `@this` parameter name.

#### 8. `git.identity` wired

`ReleaseCommand`'s committer-identity chain becomes: **configured identity → `server.CIBotIdentity` → the repository's Git config → fail.**

The CI bot outranks the repository's own Git config: commits produced by `bv` should fall back to a human's identity only as a last resort. This reverses the current order, so `ReleaseCommandTests.Release_KeepsConfiguredIdentity_WhenTheRepositoryHasOne` changes meaning and must be rewritten, alongside a new test for the configured-identity case.

#### 9. Validation split

- **Wire/shape** — schema validation, unchanged in `Buildvana.Core.Configuration`. Extend the schema so that a `nuget.feeds.*` object, if present, requires both `source` and `apiKeyEnv`; the domain feed record then has non-nullable members and the two "feed has no source / no apiKeyEnv" runtime throws disappear.
- **Selection/absence** — stays at point of use. `ResolvePushTarget(isPrerelease)` needs a runtime fact, and "no feed configured" must not fail a `bv build` in a repository that never pushes.
- The prerelease → release channel fallback moves into the factory; only the selection remains at use. Keep the channel a diagnostic mentions faithful to what the user wrote.

#### 10. Serialization and schema generation

- `BuildvanaJsonContext` in `Buildvana.Runtime` covers hook args and the domain model. `Buildvana.Runtime` stays strictly BCL.
- A new source-generated context in `Buildvana.Core.Configuration` covers the wire models.
- `tools/generate-config-schema.cs` targets `BuildvanaJsonConfig`. That it now describes a different type from the one hooks consume is the confirmation that the split is real: a JSON Schema describes a file.
- The generator should emit `default` keywords read from the domain records, replacing defaults currently stated as prose in three unlinked places. If `Buildvana.Core.JsonSchema` cannot express this cleanly, dropping it is acceptable — record the decision in this issue's thread rather than opening a follow-up.

### Usage examples

The configuration file is unchanged; this is a `buildvana.jsonc` that behaves identically before and after:

```jsonc
{
  "release": { "branches": ["^main$"], "dogfood": true },
  "dotnet": { "configuration": "Release" },
  "github": { "tokenEnv": "GITHUB_TOKEN" }
}
```

A hook reading configuration, before — one resolvable value, one that cannot be resolved at all:

```csharp
var config = hookArgs.LoadConfig();
var configuration = config.DotNet.EffectiveConfiguration;   // only correct when bv had no --configuration
var dogfood = config.Release?.Dogfood ?? true;              // duplicates bv's default; ignores --dogfood
```

After — the value the run actually used, with no default restated:

```csharp
var config = hookArgs.RuntimeInfo.Configuration;
var configuration = config.DotNet.Configuration;
var dogfood = config.Release.Dogfood;
```

Reading a credential, without touching the environment and without the name ever leaving the model:

```csharp
var feed = config.NuGet.Feeds.Release;
await PushAsync(feed.Source, feed.ApiKey);   // ApiKey resolves the variable named by feed.ApiKeyEnv
```

A repository pinning its automated-commit identity, which currently parses, validates, and does nothing:

```jsonc
{
  "git": { "identity": { "name": "Buildvana Bot", "email": "bot@buildvana.invalid" } }
}
```

### Risks

- **`Buildvana.Runtime`'s public surface changes substantially**: types move out, extension classes and loader methods are removed, new ones appear, `RuntimeInfo` gains a required member. No `**BREAKING CHANGE**:` entry is warranted — `Buildvana.Runtime` has never shipped stable — but every existing changelog entry that contradicts the new state must be amended. Two are known: the `git.identity` entry ("validated and exposed, but not yet wired to release commits") and the `Buildvana.Runtime` entry, which names `BuildvanaConfig.Load()`, `BuildvanaConfig.LoadFile()`, the well-known file-name constants, `hookArgs.LoadConfig()`, and the `Effective…` accessors — all of which change.
- **Consumers of published previews break.** Acceptable on a preview line, and the cost only grows from here.
- **Behavior change**: a hook rewriting the configuration file mid-run no longer affects what later hooks see. Intended, and the only in-repo hook that rewrites the file runs last.
- **Behavior change**: the CI bot identity now outranks the repository's Git config, so a release run on a machine with a local Git identity will use the bot's. Intended (§8); needs the changelog to say so.
- **Secrets in serialized args.** The domain model reaching `.buildvana-temp/hook-args/**` makes credential storage a security defect rather than a design preference. Needs an explicit test that no resolved secret appears in serialized hook args.
- **Cache-key change**: `ComputeVersion`'s fingerprint now covers a differently-shaped configuration source. Behaviour-preserving, but worth a test.
- No new dependencies. No performance impact expected: resolution moves from many call sites to one, run once.

### Additional information

**Out of scope, deliberately:**

- Splitting `HookArgs` / `RuntimeInfo` into wire and domain forms (see §6).
- Introducing operation-shaped command-line options such as `--no-dotnet-env`. The factory shape must not preclude them; none are added here.
- Storing any credential value in the domain model, ever.

**Acceptance criteria:**

- [ ] Wire models live in `Buildvana.Core.Configuration`; the domain model lives in `Buildvana.Runtime`; `Buildvana.Runtime` has no JSON-file, schema, or command-line knowledge and stays strictly BCL.
- [ ] Every default is a property initializer on a domain record; no default value literal exists anywhere else.
- [ ] `BuildvanaConfigFactory` is the only place that applies precedence; no consumer contains a `??` chain over configuration.
- [ ] `bv` and `ComputeVersion` both compose through that factory.
- [ ] Per-command `dotnet` args and env are resolved by the factory, including arguments forwarded after `--`; `DotNetService` contributes only base and trailing arguments.
- [ ] `RuntimeInfo.Configuration` carries the domain model; `RuntimeInfo.ConfigFile` documents that it names the source file; `HookArgs.LoadConfig()` is gone.
- [ ] Credentials are stored as variable names and resolved through extension properties, with a test asserting no resolved secret appears in serialized hook args.
- [ ] `git.identity` participates in the committer-identity chain in the order given in §8, with tests.
- [ ] Schema generation targets the wire model; `schemas/buildvana.schema.json` regenerated; `nuget.feeds.*` requires `source` and `apiKeyEnv` when present.
- [ ] This repository's own `buildvana.jsonc` updated, including its worked-example comments.
- [ ] `PublicAPI.Unshipped.txt` for `Buildvana.Runtime` updated for every removal and addition.
- [ ] `CHANGELOG.md` updated under `## Unreleased changes`, including amendment of the `git.identity` and `Buildvana.Runtime` entries.
- [ ] Docs updated: `docs/Hooks.md` (args table, configuration access, removal of `LoadConfig`), `docs/DirectoryStructure.md`, `docs/EnvironmentVariables.md` where it states defaults.
- [ ] Zero errors, zero warnings, all tests pass, `dotnet run .claude/tools/inspect.cs --gate` exits zero.

**Not deferrable.** This is the last change before the 2.1 line goes stable; anything discovered mid-implementation is handled inside the PR, per the repository's "fix it now" rule.

rdeago:

- [x] Closes #350

## Proposed changes

`BuildvanaConfig` modeled the configuration _file_; nothing modeled the _settings_, so every consumer resolved defaults and precedence privately, across the five idioms catalogued in #350. This PR gives resolution a home, as designed there:

- **Wire models** (`Buildvana.Core.Configuration`): `*JsonConfig` records faithful to the file, all members nullable (`null` = not stated in the file), plus `CommandLineOverrides` and the file discovery/loading code moved out of `Buildvana.Runtime`. The provider types are renamed to match (`BuildvanaJsonConfigProvider` and friends).
- **Domain model** (`Buildvana.Runtime`): the same record names, now the single resolved answer. Every default is a property initializer on the record that owns it; a nullable member is a domain option whose `null` has exactly one documented meaning, never "unspecified". The assembly keeps no JSON-file, schema, or command-line knowledge and stays strictly BCL.
- **One factory** (`BuildvanaConfigFactory.Create(json, commandLine)`): the only place precedence is applied — scalars flag → file → default; `dotnet` args appended `all` → per-command → forwarded after `--`; env layered by key; `release.configuration` falling back to the resolved `dotnet.configuration`; the prerelease feed falling back to the release feed. Both producers compose through it: `bv` (DI, with a real `CommandLineOverrides`) and `ComputeVersion` (task host, `null`).
- **The `DotNetService` seam**: the factory resolves each command's final args and environment; the service contributes only its base and trailing arguments. The `DotNetSettings` family dissolves, and forwarded arguments become a command-line contribution to the same resolution instead of a fourth path for the build configuration: the `-c`/`--configuration` tokens bv promotes out of the forwarded stream are also consumed by it, so `bv pack -- -c Debug` resolves and builds `Debug` without `dotnet restore` (which rejects `-c`) ever seeing the flag.
- **Hooks**: `RuntimeInfo` gains a required `Configuration` member carrying the domain model, snapshotted into each args file as it is written; `ConfigFile` is re-documented as the source-file pointer for hooks that edit the file itself; `HookArgs.LoadConfig()` and `BuildvanaConfig.Load()` are gone.
- **Credentials**: the model stores environment-variable _names_; `GetToken()` / `GetApiKey()` extension methods resolve values on demand and throw `BuildvanaRuntimeException`, translated to `BuildFailedException` at bv's call sites. A test asserts no resolved secret ever appears in serialized hook args. One deviation from §7: extension _methods_, not extension _properties_ — a getter that reads process state and throws on absence would betray property semantics, while the structural cannot-be-serialized guarantee is identical.
- **Schema**: the generator gains `default` emission, annotated from a domain-model instance, so the schema documents exactly what an omitted setting resolves to (`required` needed no generator code — System.Text.Json already emits it for C# `required` members, now locked by a test). `nuget.feeds.*` requires `source` + `apiKeyEnv` and `git.identity` requires `name` + `email`, so a half-written section fails at configuration load and the domain records carry non-nullable members. Blank is never a value: required strings carry `minLength`/`pattern` constraints enforced at load (BV1106/BV1107, with file and line), optional strings treat a stated blank as unstated, and a blank CLI option value is rejected like a missing one. `schemas/buildvana.schema.json` regenerated.
- **`git.identity` is live**: `bv release` resolves its committer identity as configured identity → CI bot identity → repository Git config → fail. The bot now outranks whatever a CI checkout left in the repository's config (recorded as a behavior change in the changelog); a configured identity outranks both.
- **Tests**: 904 passing; new coverage includes the identity chain end to end, the secrets-never-serialized guarantee, BV1104 on half-written sections, blank-value rejection at each input layer, the feed fallback, and version-cache invalidation when the configuration file changes between builds. `dotnet run .claude/tools/inspect.cs --gate` exits zero.

Docs (`Hooks.md`, `DirectoryStructure.md`, `EnvironmentVariables.md`), `PublicAPI.Unshipped.txt`, the changelog (including the amendments §Risks called for), and this repository's own `buildvana.jsonc` are updated per the acceptance criteria.

## Additional changes

- `EnvVarHelper.Require`'s message reworded to the standard adopted for `EnvironmentVariables.GetRequired` ("Required environment variable X is missing or empty."), so the one remaining non-configuration env read (`GITHUB_OUTPUT`) fails with the same words as everything else.
- The committer-identity failure message no longer ends with "before running this task" (Cake-era phrasing).
- `UnknownServerAdapter.CIBotIdentity` now answers `null` — a local build has an honest answer to that question — instead of failing the build; the class doc, already false for `PushUsername`/`PushPassword`, corrected to match.
- `CHANGELOG.md`: dropped the stale claim that the computed version is reported at `notice:` level (the task reports it at detail level).
- `docs/Hooks.md`: the hook-args logging level corrected to trace (the doc said detail; the code says trace).
- `docs/ConfigurationFiles.md` → `docs/SdkConfigurationFiles.md`, with a new `docs/BuildvanaToolConfiguration.md` TODO stub for `buildvana.json`; every reference re-pointed to whichever of the two it actually meant.
- `buildvana.jsonc`: four worked-example comments re-synced to the schema descriptions the file's intro promises they mirror.
- `Buildvana.slnx`: stale file entries fixed (`.globalconfig`, `docs/Diagnostics.md`, `THIRD-PARTY-NOTICES` without extension, `NuGet.config` casing) and the real documentation and configuration file sets listed.
- `CommandParameters.Forwarded`: the ReSharper-suppression justification named the wrong reader (the `BuildPipeline` constructor; it is `CommandLineOverridesParser`), and the member docs overstated "verbatim" forwarding now that bv consumes the configuration tokens out of the stream; both corrected.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [x] Other

"Other" covers the restructuring itself: `Buildvana.Runtime`'s public surface changes substantially (types moved out, loaders removed, extension methods and a required member added), which the "Refactor" box excludes by its no-public-API-changes parenthetical.

### Breaking changes

- [ ] Yes
- [x] No

`Buildvana.Runtime` has never shipped stable, and the identity-precedence change is recorded as a behavior change on the preview line; nothing stable breaks.

## Checklist

- **For code changes only:**


src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
```diff
@@ -0,0 +1,175 @@
...
+    private static DotNetConfig ComposeDotNet(DotNetJsonConfig? json, CommandLineOverrides? commandLine)
+    {
+        var defaults = new DotNetConfig();
+        var all = ComposeInvocation(json?.All, common: null, forwardedArgs: null);
+        var forwardedArgs = commandLine?.ForwardedArgs;
+        return new DotNetConfig
+        {
+            Configuration = commandLine?.Configuration ?? json?.Configuration ?? defaults.Configuration,
+            All = all,
+            Restore = ComposeInvocation(json?.Restore, all, forwardedArgs),
```
rdeago, line 61: **Correctness.** The `-c` promotion and this fold are on a collision course: the forwarded arguments that `CommandLineOverridesParser` promotes into `CommandLineOverrides.Configuration` are also folded verbatim into `Restore.Args` here, and `dotnet restore` rejects `-c` with `MSBUILD : error MSB1001: Unknown switch` (confirmed empirically on SDK 10.0.303). Every pipeline run starts at the Restore step, so the exact syntax this PR advertises — `bv pack -- -c Debug`, per the new CHANGELOG bullet and the comment in `CommandLineOverridesParser` — fails before the override ever takes effect. `Create_ForwardedArgs_AppendToPipelineCommandsOnly` even asserts `Restore.Args` ends with `-c|Debug`, baking the failure in as expected behavior.

To be fair, the breakage itself is not new: on `main`, forwarded args already reached `dotnet restore`, so `-- -c Debug` already died with MSB1001. But this PR is the one that documents the syntax as working and promotes the flag into bv's own configuration, so it has to make it actually work. Since bv now interprets and owns the forwarded `-c`/`--configuration`, the natural fix is to strip the recognized tokens from the forwarded stream once promoted — bv already injects the resolved configuration explicitly where it belongs, which also removes the two-writers situation for build/test/pack.

src/Buildvana.Tool/CommandLine/CommandLineOverridesParser.cs
```diff
@@ -0,0 +1,44 @@
...
+        Guard.IsNotNull(parameters);
+        var reader = new CliOptionReader(parameters.Options);
+
+        // A configuration stated among the forwarded arguments decides the actual build, so bv's own view
+        // must agree with it: `bv pack -- -c Debug` resolves bv's configuration to Debug. The reader works
+        // on a copy of the tokens, so the forwarded arguments themselves still reach `dotnet` verbatim.
+        var forwardedReader = new CliOptionReader(parameters.Forwarded);
+        return new CommandLineOverrides
+        {
+            Configuration = reader.ReadValue("--configuration", "-c") ?? forwardedReader.ReadValue("--configuration", "-c"),
```
rdeago: **Correctness.** Reading the forwarded stream with bv's own option grammar breaks the documented pass-through contract (the pre-existing CHANGELOG bullet: everything after the first `--` is passed through verbatim, `bv` no longer parses or validates it, and malformed forwarded arguments produce an error from `dotnet`, not from `bv`):

- A trailing `-c` — e.g. `bv test -- --treenode-filter -c`, where `-c` is the _value_ of the MTP option — makes `CliOptionReader.ReadValue` throw `Option '-c' requires a value.` during DI resolution of `BuildvanaConfig`, before any child process starts.
- `bv test -- --some-opt -c X`, where `-c` was `--some-opt`'s value, silently captures `X` as bv's build configuration.

`CliOptionReader` matches tokens positionally, with no notion of "this token is another option's value" — fine for bv's own options, not for an arbitrary child grammar. Whatever shape the fix to the Restore collision takes (see the factory comment), it should settle this contract question explicitly: either bv owns `-c`/`--configuration` in the forwarded stream (then strip them after promotion and amend the verbatim-pass-through bullet), or it does not (then don't read them).

CHANGELOG.md
```diff
@@ -97,6 +98,7 @@ See the Nerdbank.GitVersioning removal entry under _Changes to existing features
   - delete `version.json` and remove `nbgv` from your tool manifest, if present.
 - **BREAKING CHANGE**: The `UseNerdbankGitVersioning` property has been renamed to `UseVersioning`. The old name still works as an alias (when `UseVersioning` itself is not set), but raises warning BVSDK2001 and will be removed in a future version.
 - **BREAKING CHANGE**: `version.json` features without a native equivalent are dropped: `pathFilters` is not supported (the version height is always computed from the whole history of the version line), nested `version.json` files are not supported (`VERSION` lives at the repository root only), and the NuGet package version scheme is fixed at SemVer 2.0.
+- A `-c`/`--configuration` argument forwarded after `--` now also drives `bv`'s own view of the build configuration: `bv pack -- -c Debug` builds, reports, and locates artifacts for `Debug`, where it previously let the forwarded flag decide the actual build while `bv` itself kept resolving (and reporting) `Release`.
```
rdeago: Two problems with this bullet, both downstream of the forwarded-`-c` findings: it documents a command line that currently fails at the Restore step (`bv pack -- -c Debug` → MSB1001, see the factory comment), and the "previously let the forwarded flag decide the actual build" claim is inaccurate — previously the same command line also died at Restore, so the forwarded `-c` never decided any build either. Needs rewording once the mechanics are settled.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    // The schema requires source and apiKeyEnv whenever a feed is stated, so a wire feed that reaches the
+    // factory carries both.
+    private static NuGetFeedConfig? ComposeFeed(NuGetFeedJsonConfig? json)
+        => json is null
+            ? null
+            : new NuGetFeedConfig
+            {
+                Source = json.Source!,
```
rdeago, line 157: **Correctness (blank-vs-absent gap).** `required` checks presence and `"type": "string"` accepts `""`, so `"source": ""` / `"apiKeyEnv": ""` sail through schema validation and these `!` mappings. Pre-PR, `DotNetSettings.ResolvePushTarget` rejected blank members with an actionable error ("NuGet feed 'nuget.feeds.release' has no source. Set its source in the configuration file.") — those guards and their tests were deleted with no replacement. Now a blank `apiKeyEnv` surfaces at push time as `Required environment variable  is missing or empty.` (naming no variable), and a blank `source` becomes `dotnet nuget push --source ""` with NuGet's raw error. The PR's "a half-written section fails at configuration load" guarantee holds only for absent members.

One fix covers the whole family (also the `git.identity`, configuration, and `tokenEnv` comments): make "a blank string is not a value" a single rule at one layer — either the generator emits `minLength: 1` for required strings (schema and validator are repo-owned, so the failure is reported with file/line like any other schema error), or the factory rejects stated-but-blank members with the actionable message the old code had.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
```diff
@@ -0,0 +1,175 @@
...
+        };
+    }
+
+    // The schema requires name and email whenever git.identity is stated, so a wire identity that reaches
+    // the factory carries both.
+    private static GitConfig ComposeGit(GitJsonConfig? json)
+        => new()
+        {
+            Identity = json?.Identity is { } identity
+                ? new GitIdentityConfig { Name = identity.Name!, Email = identity.Email! }
```
rdeago, line 177: **Correctness (blank-vs-absent gap, worst member of the family).** `"name": ""` / `"email": ""` pass schema validation and are stamped into `GitIdentityConfig` verbatim. `bv release` then _writes_ `user.name` / `user.email` = `""` into the repository's `.git/config` — the `GitService.CommitterIdentity` setter persists at the Local level, and nothing rolls it back on failure. LibGit2Sharp's `BuildSignature` treats empty as missing and returns null, so the run crashes with a `NullReferenceException` at `ReleaseCommand.cs:87` when reporting the identity — an internal crash instead of the config-load failure the CHANGELOG advertises — and the polluted `.git/config` stays behind. Whitespace-only values get further: they survive `BuildSignature` and fail at the first release commit, after the full verification build. Same family fix as the feed comment above.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
```diff
@@ -0,0 +1,175 @@
...
+    // the push's argument surface is not the pipeline's. The All member itself stays as written, for
+    // consumers that want the common tier alone.
+    private static DotNetConfig ComposeDotNet(DotNetJsonConfig? json, CommandLineOverrides? commandLine)
+    {
+        var defaults = new DotNetConfig();
+        var all = ComposeInvocation(json?.All, common: null, forwardedArgs: null);
+        var forwardedArgs = commandLine?.ForwardedArgs;
+        return new DotNetConfig
+        {
+            Configuration = commandLine?.Configuration ?? json?.Configuration ?? defaults.Configuration,
```
rdeago: **Correctness (blank-vs-absent gap, third member).** A blank configuration — `"configuration": ""` in the file, or the inline forms `-c=` / `--configuration=`, which `CliOptionReader.TryMatchInline` returns as `""` (including from forwarded args, a path new in this PR) — survives these `??` chains as the resolved configuration and dies as `Guard.IsNotNullOrEmpty`'s `ArgumentException` in `BuildPipeline.RunAsync`: an internal-parameter crash, not a configuration diagnostic. The file-sourced variant crashed equivalently pre-PR, but the factory is now the declared single composition point — exactly the right place to diagnose it. Same family fix as the feed comment above.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
```diff
@@ -0,0 +1,175 @@
...
+                ApiKeyEnv = json.ApiKeyEnv!,
+            };
+
+    // A blank tokenEnv cannot name a variable, so it counts as not stated at all.
+    private static GitHubConfig ComposeGitHub(GitHubJsonConfig? json)
+    {
+        var defaults = new GitHubConfig();
+        return new GitHubConfig
+        {
+            TokenEnv = json?.TokenEnv is { Length: > 0 } name ? name : defaults.TokenEnv,
```
rdeago: **Minor (same family).** The comment says "a blank tokenEnv cannot name a variable, so it counts as not stated at all", but `{ Length: > 0 }` only catches `""`: a whitespace-only `" "` is kept as the variable name — no `GITHUB_TOKEN` fallback — and fails at `GetToken()` with a message whose variable name is invisible whitespace. `NormalizeBlankToNull` a few lines up already defines blank as `IsNullOrWhiteSpace`; reuse it here so the file carries one definition of "blank".

src/Buildvana.Tool/Services/DotNetService.cs
```diff
@@ -241,10 +229,20 @@ public async Task NuGetPushAllAsync(string artifactsPath, CancellationToken canc
...
         }
 
-        var target = _settings.ResolvePushTarget(_version.IsPrerelease);
+        var feed = ResolvePushFeed(_config.NuGet.Feeds, _version.IsPrerelease);
+        string apiKey;
+        try
+        {
+            apiKey = feed.GetApiKey();
+        }
+        catch (BuildvanaRuntimeException e)
```
rdeago: **Minor (duplication).** This try/catch translating `BuildvanaRuntimeException` into `BuildFailedException` is byte-identical to the one around `GetToken()` in the `GitHubServerAdapter` constructor. Routing secret access through the Runtime accessors is clearly the right design (one code path shared with hooks), so the translation wrap is the part to share: a small Tool-side helper — e.g. `static T Translate<T>(Func<T> accessor)` — keeps it in one place for the third secret accessor that will inevitably arrive.

src/Buildvana.Core.JsonSchema/JsonSchemaGenerator.cs
```diff
@@ -286,4 +309,63 @@ private static void RemoveNullFromEnum(JsonObject schema)
...
+            }
+
+            var schemaProperty = FindPropertyByJsonName(schemaType, jsonName, options);
+            var noDefault = schemaProperty?.IsDefined(typeof(JsonSchemaNoDefaultAttribute), inherit: true) ?? false;
+            if (schemaProperty is null || noDefault)
+            {
+                continue;
+            }
+
+            var value = FindPropertyByJsonName(defaults.GetType(), jsonName, options)?.GetValue(defaults);
```
rdeago: **Robustness.** `FindPropertyByJsonName(...)?.GetValue(defaults)` conflates "the defaults type has no property with this JSON name" with "the default is null", so a rename or a lost `[JsonPropertyName]` on either side silently drops a whole section's defaults — and the wire and domain types must mirror spellings by hand (`"dotnet"`/`"nuget"`/`"github"` are pinned on both `BuildvanaJsonConfig` and `BuildvanaConfig` for exactly this reason). CI's schema check does flag the resulting diff, but the documented remedy is `--update` and commit, which bakes the loss in without any signal from the generator itself.

Since the code can tell the two cases apart, make the miss loud: when the defaults type has no matching property and the schema property is neither opted out via `[JsonSchemaNoDefault]` nor injected via `[JsonAllowedKeys]`, throw instead of skipping. Prerequisite: `BuildvanaJsonConfig.Schema` (`$schema`, wire-only by design, legitimately without a domain counterpart) needs the opt-out first.

src/Buildvana.Tool/Subcommands/CleanCommand.cs
```diff
@@ -4,18 +4,19 @@
...
 
 [ImplementsCommand("clean")]
 [Description("Remove all build artifacts, intermediate output, and temporary files. Like 'dotnet clean', but more aggressive.")]
-internal sealed class CleanCommand(BuildPipeline pipeline) : IBvCommand
+internal sealed class CleanCommand(BuildPipeline pipeline, BuildvanaConfig config) : IBvCommand
 {
     public async Task<int> ExecuteAsync(CancellationToken cancellationToken)
     {
-        await pipeline.RunThroughAsync(BuildStep.Clean, cancellationToken: cancellationToken).ConfigureAwait(false);
+        await pipeline.RunThroughAsync(BuildStep.Clean, config.DotNet.Configuration, cancellationToken).ConfigureAwait(false);
```
rdeago: **Judgment call (simplification).** Five commands now inject `BuildvanaConfig` solely to hand `config.DotNet.Configuration` to the pipeline — including Clean and Restore, whose steps the pipeline's own doc says ignore the value (it is even guarded non-empty on a path that never uses it). Consider letting `BuildPipeline` take `BuildvanaConfig` and resolve the default itself — same DI graph, no cycle (checked: the config factory depends only on the JSON provider and `CommandLineOverrides`) — leaving only `ReleaseCommand` to pass an explicit `config.Release.Configuration`. Trade-off acknowledged: it reintroduces the optional parameter this PR deliberately removed. But as it stands, "pass a required configuration to a step documented to ignore it" reads like a bug at every new call site.

src/Buildvana.Tool/Services/DotNetService.cs
```diff
@@ -86,15 +85,13 @@ public Task RestoreSolutionAsync(SolutionContext solution, IReadOnlyList<string>
     /// </summary>
     /// <param name="solution">The solution to build.</param>
     /// <param name="configuration">The MSBuild configuration to build.</param>
-    /// <param name="forwardedArgs">Extra arguments to forward verbatim to the <c>dotnet</c> invocation.</param>
     /// <param name="restore"><see langword="true"/> to restore NuGet packages before building, <see langword="false"/> otherwise.</param>
     /// <param name="cancellationToken">A token that, when signalled, terminates the spawned <c>dotnet</c> child process.</param>
     /// <returns>A <see cref="Task"/> representing the ongoing operation.</returns>
-    public Task BuildSolutionAsync(SolutionContext solution, string configuration, IReadOnlyList<string> forwardedArgs, bool restore, CancellationToken cancellationToken = default)
+    public Task BuildSolutionAsync(SolutionContext solution, string configuration, bool restore, CancellationToken cancellationToken = default)
```
rdeago: **Style (line length).** Modified declarations exceed the 120-character declaration threshold and need wrapping: line 91 (143 chars), line 122 (160), line 193 (154). Line 307 (143 chars) exceeds the general 140-character limit.

src/Buildvana.Tool/Build/BuildPipeline.cs
```diff
@@ -25,55 +23,47 @@ internal sealed class BuildPipeline
...
         _hookRunner = hookRunner;
-        _forwardedArgs = parameters.Forwarded;
     }
 
     /// <summary>
     /// Runs the pipeline from <see cref="BuildStep.Clean"/> through <paramref name="last"/>, inclusive.
     /// </summary>
     /// <param name="last">The last step to run.</param>
-    /// <param name="configuration">The MSBuild configuration to build, or <see langword="null"/> to use the configured default (<see cref="DotNetSettings.Configuration"/>).</param>
+    /// <param name="configuration">The MSBuild configuration to build (ignored by <see cref="BuildStep.Clean"/> and <see cref="BuildStep.Restore"/>).</param>
```
rdeago: **Style (line length).** The new `<param name="configuration">` doc lines at 52, 63, and 80 are 158 characters (the 140 limit covers XML doc comments), and the modified `RunRangeAsync` declaration at line 66 is 137 characters (declaration threshold is 120). Note these lines change anyway if the pipeline-resolves-its-own-default suggestion in `CleanCommand` is taken.

src/Buildvana.Core.JsonSchema/JsonSchemaGenerator.cs
```diff
@@ -50,8 +65,11 @@ public static JsonNode Generate<T>(JsonSerializerOptions options, string? title
     /// <param name="options">The serializer options that govern property naming, enum formatting, and so on.</param>
     /// <param name="title">The schema title; when omitted, the type's <see cref="JsonSchemaTitleAttribute"/>
     /// (if any) supplies it. Useful when the type cannot carry the attribute.</param>
+    /// <param name="defaults">An instance whose property values become <c>default</c> annotations in the
+    /// schema, or <see langword="null"/> to emit none. Matched to schema properties by JSON name, so it does
+    /// not have to be of the described type; see the remarks.</param>
     /// <returns>The schema as a <see cref="JsonNode"/>.</returns>
-    public static JsonNode Generate(Type type, JsonSerializerOptions options, string? title = null)
+    public static JsonNode Generate(Type type, JsonSerializerOptions options, string? title = null, object? defaults = null)
```
rdeago: **Style (line length).** The modified `Generate` declaration (124 chars) and the new `ApplyDefaults` declaration at line 316 (121 chars) exceed the 120-character declaration threshold and need wrapping.

tests/Buildvana.Core.JsonSchema.Tests/DefaultsSchemaSample.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (C) Tenacom and Contributors. Licensed under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using System.Diagnostics.CodeAnalysis;
+using Buildvana.Core.JsonSchema;
+
+// The type the schema is generated FROM in the defaults-emission tests: all-nullable, wire-model style.
+// Its defaults deliberately live on a different type (DefaultsValuesSample), so the tests prove that
+// matching is by JSON name, not by type identity.
+[SuppressMessage("Performance", "CA1812:Avoid uninstantiated internal classes", Justification = "Reflected over by the schema generator under test; never instantiated.")]
```
rdeago: **Style (line length).** This 170-character `[SuppressMessage]` line — and its twins in `DefaultsSchemaSection.cs:7` and `RequiredSample.cs:8` — exceeds the 140-character limit. Wrapping the attribute arguments brings all three within it.

rdeago (commented):
Review of the wire/domain configuration split. The architecture does what #350 asked for — one factory owning precedence, defaults on the domain records, strictly-BCL Runtime preserved — and the migration of the deleted validation/defaulting logic was audited member-by-member: defaults all moved faithfully, the multiple-config-file guard moved with its coverage, the prerelease→release feed fallback is intact, and the deleted test scenarios are re-covered by the new factory/provider tests. Conventions all check out too (BOMs, extension-block template, no non-BCL references in Runtime, env-var test parallelism rationale).

Two correctness clusters need fixing in this PR, plus smaller items — all anchored inline:

**1. Forwarded `-c`/`--configuration` (two comments + CHANGELOG).** The advertised `bv pack -- -c Debug` fails at the Restore step: the forwarded `-c` is folded into `Restore.Args` and `dotnet restore` rejects it with MSB1001 (verified empirically). The breakage predates this PR, but this PR documents the syntax as working and bakes it into a test. Separately, parsing the forwarded stream breaks the documented "passed through verbatim, never parsed by bv" contract: a forwarded token that merely _looks_ like `-c` (e.g. as another option's value) is captured or, in trailing position, aborts the run with a bv error. Both point at one decision: if bv owns `-c`/`--configuration` in the forwarded stream, strip them after promotion and amend the CHANGELOG contract bullet; if it doesn't, don't read them.

**2. Blank-vs-absent gap in the new load-time guarantee (four comments).** `required` checks presence and `"type": "string"` accepts `""`, so the "a half-written section fails at configuration load" guarantee holds only for _absent_ members. Blank values regress diagnostics the old code had: `""` feed source/apiKeyEnv fail cryptically at push time (the old actionable "has no source" guards were deleted with their tests), blank `git.identity` members are persisted into `.git/config` and crash with an NRE (or fail post-build for whitespace), and a blank configuration dies as a `Guard` `ArgumentException`. One rule at one layer fixes the family — generator-emitted `minLength: 1` for required strings, or factory-side stated-but-blank rejection — plus a one-line `IsNullOrWhiteSpace` consistency fix for `tokenEnv`.

**Smaller items:** a silent failure mode in the schema generator's defaults matching (name mismatch drops a section's defaults; make the miss throw), the duplicated `BuildvanaRuntimeException`→`BuildFailedException` translation block (share a tiny Tool-side helper; the Runtime-accessor routing itself is right), a judgment call on threading `config.DotNet.Configuration` through five commands including two that ignore it, and line-length violations on modified lines in four places.

Also examined and deliberately **not** flagged, as they turned out to be intentional design: the settings classes still parsing flags whose resolved values come from the overrides parser (documented attribute-carrier mechanism), `DotNetConfig.All` (hooks-contract surface per the completeness principle), the factory's per-call `new` of defaults records (documented single-source-of-defaults idiom), and the schema generator's hand-rolled name resolution (no serializer disagreement is reachable with these models, and `JsonTypeInfo` wouldn't remove the reflection dependency).

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
rdeago: Done in 8b6ddc9, as suggested: the parser reads the forwarded stream through the same `CliOptionReader` that promotes the tokens — reading consumes — and publishes `Remaining` as `ForwardedArgs`, so promotion and stripping can never disagree and `dotnet restore` never sees `-c`. bv injects the resolved configuration itself, in the form each command accepts. `Create_ForwardedArgs_AppendToPipelineCommandsOnly` now folds a neutral `-m:8`, and new parser tests pin promotion + stripping for both spellings and both token forms.

src/Buildvana.Tool/CommandLine/CommandLineOverridesParser.cs
rdeago: Settled as ownership, in 8b6ddc9: bv owns these two option names wherever they appear after `--`. That is the only choice that leaves the pipeline configurable from the command line at all — the pipeline commands no longer parse `-c` in front of `--`, and no forwarded spelling suits all four commands (`dotnet restore` rejects `-c`; MTP-mode `dotnet test` rejects `-p:`). The recognized tokens are stripped after promotion; the trailing-`-c` error and the captured-value case are now defined, documented costs — the changelog states both, and `Parse_Throws_OnTrailingForwardedConfiguration` pins the error case.

CHANGELOG.md
rdeago: Done in 8b6ddc9: this bullet is folded into the breaking-change `-c` bullet, which now states the final behavior only — recognized after `--`, drives bv's own resolution, stripped from what reaches `dotnet`, with the two ownership consequences spelled out — and makes no claim about what the forwarded flag previously decided. The verbatim-pass-through bullet gained its one exception, pointing at the merged bullet.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
rdeago: Done in 4641a4d, via the schema layer: the generator emits `minLength: 1` plus `pattern: "\S"` for required string properties (the pattern closes the whitespace-only hole minLength leaves), the validator enforces both and reports the new BV1106/BV1107 with file and line, and the committed schema is regenerated. A blank `source`/`apiKeyEnv` now fails at configuration load exactly like an absent one, so the `!` mappings here are backed by the schema again. The validator reports at most one string error per value, so a blank member states one mistake once.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
rdeago: Done in 4641a4d, same mechanism as the feeds: blank and whitespace-only `name`/`email` now fail at configuration load (BV1106 for empty, BV1107 for all-whitespace, with file and line), so nothing blank can reach `GitService.CommitterIdentity` from configuration — no `.git/config` pollution, no `BuildSignature` null, no post-build failure. `Config_WhitespaceIdentityName_ReportsBV1107` covers the worst member of the family directly.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
rdeago: Done in 4641a4d, on both input paths: `CliOptionReader` now rejects a blank or whitespace value (`-c=`, `-c ""`, inline or space-separated, forwarded stream included) with the existing "requires a value" error, so overrides can never carry blank; and the factory folds a blank file-side configuration to null, so the next precedence tier applies instead of the `Guard` crash. Covered by `Create_Configuration_BlankCountsAsNotStated` and the new reader tests.

src/Buildvana.Core.Configuration/BuildvanaConfigFactory.cs
rdeago: Done in 4641a4d: `NormalizeBlankToNull` is now the factory's single definition of blank, applied to `tokenEnv` and the other optional strings (`dotnet.configuration`, `release.configuration`, `versioning.prereleaseTag`). A whitespace-only name falls back to `GITHUB_TOKEN`; the test now covers it.

src/Buildvana.Core.JsonSchema/JsonSchemaGenerator.cs
rdeago: Done in 73a4723: the defaults-side lookup now throws `InvalidOperationException` naming the defaults type, the JSON name, and the `[JsonSchemaNoDefault]` opt-out; `BuildvanaJsonConfig.Schema` carries the opt-out as of the same commit. The schema-type-side null (the `[JsonAllowedKeys]` fall-through) stays a silent skip, as does a null default value. `Generate_WithDefaults_ThrowsOnUnmatchedSchemaProperty` pins the throw, and the schema tool's check mode proves the real model pair still generates clean.

src/Buildvana.Tool/Services/DotNetService.cs
rdeago: Done in bb189cf: both sites now go through `RuntimeAccess.Translate(Func<T>)` (new, in `Buildvana.Tool/Utilities`) — `Translate(feed.GetApiKey)` here, `Translate(...GitHub.GetToken)` in the adapter constructor.

src/Buildvana.Tool/Subcommands/CleanCommand.cs
rdeago: Taken, in d828773: `BuildPipeline` now injects `BuildvanaConfig` and defaults a null `configuration` parameter to the resolved `dotnet.configuration`; the five commands dropped their config injection, and `ReleaseCommand` alone passes `release.configuration` explicitly (it needs the value anyway to locate artifacts). The optional parameter returns as you noted — judged less misleading than requiring a value two of five call sites cannot use.

src/Buildvana.Tool/Services/DotNetService.cs
rdeago: Done in 2c97e54, with the whole-file sweep the line-length rule asks for: besides the three declarations and the message line, the Pack summary line and the `RunDotNetAsync` returns line were also over 140 and got wrapped.

src/Buildvana.Tool/Build/BuildPipeline.cs
rdeago: Done — absorbed by d828773, as you anticipated: the pipeline refactor rewrote all three declarations and their doc blocks, and the file now has no line over either limit.

src/Buildvana.Core.JsonSchema/JsonSchemaGenerator.cs
rdeago: Done in 00af27d.

tests/Buildvana.Core.JsonSchema.Tests/DefaultsSchemaSample.cs
rdeago: Done in fed25d7 — including a fourth twin of the same attribute in `GeneratorSample.cs:10`.

rdeago:
All 14 threads addressed and resolved — nine commits, one per finding, shared-fix threads sharing one:

- 8b6ddc9 — bv owns `-c`/`--configuration` in the forwarded stream and strips it after promotion; the changelog's `-c` story consolidated into one final-state bullet, the verbatim-pass-through bullet carrying its one exception (3 threads).
- 4641a4d — blank-is-not-a-value family: `minLength`/`pattern` on required strings (BV1106/BV1107 at load, with file and line), factory folds blank optional strings to unstated, reader rejects blank option values (4 threads).
- 73a4723 — schema generator: a defaults-side miss now throws; `$schema` opted out via `[JsonSchemaNoDefault]`.
- bb189cf — shared `RuntimeAccess.Translate` for the Runtime-to-`BuildFailedException` translation.
- d828773 — `BuildPipeline` resolves its own default configuration; the five commands drop their config injection (this also rewrote every line the BuildPipeline line-length thread flagged).
- 2c97e54, 00af27d, fed25d7 — line-length wraps in DotNetService, JsonSchemaGenerator, and the schema-test samples (a fourth `SuppressMessage` twin in `GeneratorSample.cs` included), each with its whole-file sweep.

One small out-of-scope fix on the way, recorded in the PR description's Additional changes: a1c6190 corrects `CommandParameters.Forwarded`'s stale ReSharper justification (it named the `BuildPipeline` constructor; the reader is `CommandLineOverridesParser`) and its now-overstated "verbatim" wording.

Gate is green (`bv pack` + inspectcode, zero findings), 904 tests passing, schema check clean. The PR description is updated to describe the branch as it now stands.

@rdeago please take another look.

rdeago:
Re-reviewed the nine response commits against every thread. All 14 findings are correctly resolved, and the new diff introduces no new problems. Verified in particular:

- **8b6ddc9** — promotion and stripping share one `CliOptionReader` pass (`Remaining` preserves order), the read runs unconditionally so the stream is stripped even when bv's own options win, `dotnet restore` never sees `-c`, and the CHANGELOG now tells one coherent story: the pass-through bullet carries its single exception, and the consolidated `-c` bullet states the final behavior with both ownership costs spelled out honestly.
- **4641a4d** — the non-blank constraint lives in exactly one layer per input kind: `minLength: 1` + `pattern: "\S"` on required strings only (pinned by tests on all three sides: required string constrained, optional string and required non-string untouched), BV1106/BV1107 wired end to end (validator, provider mapping, docs table, regenerated schema), one `NormalizeBlankToNull` for every optional string in the factory, and the option reader rejecting blank values in both token forms. The code-point `minLength` counting and the pattern timeout guard are nice touches.
- **73a4723** — the schema-side skip (`[JsonSchemaNoDefault]`, `[JsonAllowedKeys]` fall-through) stays ahead of the new defaults-side throw, so the `$schema` opt-out and dictionary keys still pass while a genuine miss now fails with an actionable message.
- **bb189cf / d828773 / a1c6190** — as described in the replies; the pipeline's `??=` default plus the factory's blank normalization means the surviving `Guard` is now genuinely unreachable from configuration input.
- **Line lengths** — measured every touched file: no line over 140 and no declaration over 120 anywhere in the nine commits. The only over-limit lines left in touched files (`JsonSchemaValidator.cs:122`, `CommandParameters.cs:13–14`, `JsonSchemaValidatorTests.cs:162`) all pre-date the branch on `main`, so nothing this PR added or modified is out of compliance; a wrap-only commit for those four would complete the whole-file sweep the style guide suggests, but that's optional cleanup, not a gate.

Codecov (patch and project) and CodeQL are green on `fed25d7`. This PR is ready to merge.

